PR #2, opened 2017-09-06
Implement GetOutKeystream()

sergeyfrolov:


u_conn.go
```diff
@@ -435,3 +435,16 @@ func (uconn *UConn) MarshalClientHello() error {
 	hello.Raw = helloBuffer.Bytes()
 	return nil
 }
+
+// get current state of cipher and encrypt zeros to get keystream
+func (uconn *UConn) GetOutKeystream(length int) (keystream []byte, err error) {
+	zeros := make([]byte, length)
+
+	if servConnCipher, ok := uconn.out.cipher.(cipher.AEAD); ok {
+		keystream = servConnCipher.Seal(nil, uconn.out.seq[:], zeros, nil)
```
ewust: Should comment that (we assume) Seal() does not update any internal state (as in, it can be called multiple times and get the same result). This assumption may not be universal, so someone extending it to say, CBC mode might want to know that.

sergeyfrolov: Done

PR #3, opened 2018-02-13
Add option to omit SNI

sergeyfrolov:
Not sure whether this feature is ready to be used as it is, since certificate together with name and is still easily parsed from server response.
Omitting SNI would be more useful if we had designed auth protocol over fake session tickets, which would skip the certificate as well.


sergeyfrolov:
Doesn't seem like there's an interest in this.
Which is not surprising: your tool should send SNI, because [everybody does](https://tlsfingerprint.io/find/extension/0000) and your tool has to not stick out.

PR #6, opened 2018-06-19
Refactor: enable flexible external configuration

sergeyfrolov:
Most deletions are due to removal of Android test traces.


u_common.go
```diff
@@ -74,49 +74,68 @@ const (
...
-	HelloAndroid_Auto        ClientHelloID = ClientHelloID{helloAndroid, helloAutoVers}
-	HelloAndroid_6_0_Browser ClientHelloID = ClientHelloID{helloAndroid, 23}
-	HelloAndroid_5_1_Browser ClientHelloID = ClientHelloID{helloAndroid, 22}
+	HelloChrome_Auto = HelloChrome_62
+	HelloChrome_58   = ClientHelloID{helloChrome, 58}
+	HelloChrome_62   = ClientHelloID{helloChrome, 62}
 )
 
+// based on spec's GreaseStyle, GREASE_PLACEHOLDER may be replaced by another GREASE value
+const GREASE_PLACEHOLDER = 0x0a0a
```
willscott: is this gonna have to be updated every 6 weeks or whatever?

sergeyfrolov: why would it have to be updated every 6 weeks or whatever?

willscott: I see, this is the template pattern that stays stable. makes sense.

handshake_messages.go
```diff
@@ -20,7 +20,7 @@ type clientHelloMsg struct {
 	serverName                   string
 	ocspStapling                 bool
 	scts                         bool
-	ems                          bool
+	ems                          bool // [UTLS]
```
bemasc: This comment could use some expansion.

u_common.go
```diff
@@ -74,49 +74,68 @@ const (
 	helloRandomizedNoALPN
 )
 
+type ClientHelloSpec struct {
+	ClientRandom       []byte
```
bemasc: Can this be nil?

u_common.go
```diff
@@ -74,49 +74,68 @@ const (
...
+	CipherSuites       []uint16       // nil => default
+	CompressionMethods []uint8        // nil => no compression
+	Extensions         []TLSExtension // nil => no extensions
+
+	// GreaseStyle: currently only Chrome(GREASE depends on ClientRandom)
```
bemasc: Does GREASE still depend on ClientRandom?

handshake_messages.go
sergeyfrolov: Throughout the standard files, which we are not supposed to change too much, I was leaving `[UTLS]` comment to explicitly state what we added.
I can switch to something like `[Added by UTLS]`, if that's better

u_common.go
sergeyfrolov: Yes, but now that I thought of it, we probably shouldn't have it in spec at all, as it's supposed to be a generic descriptor of configuration.
Will remove.

u_common.go
sergeyfrolov: Good point, forgot to update this.

handshake_messages.go
bemasc: I understood that it was marked as added.  I was hoping you would expand the comment to explain _why_ it was added.

sergeyfrolov: Alright, `[UTLS]` -> `[UTLS] extension implemented due to its prevalence`

PR #8, opened 2018-07-28
Fix statefulness

sergeyfrolov:
With last refactoring, that enabled flexible external configs(and ultimately enabled effortless use of auto-generated code for any fingerprint, including recently integrated [iOS 11](https://tlsfingerprint.io/id/a9ecd7afe6b0e4a4) one), we introduced a regression:
*Various connections, using same ClientHelloSpec may share state.*

To provide more info on what state was erroneously shared, here are some examples:
 - Shared slice with curves for `SupportedCurvesExtension` - should only change, if manually changed by user, which I don't think anyone does.
 - List of ALPN values in extensions might be realistically be changed by user(albeit, it shouldn't be! danger! Chrome-like fingerprint without `h2` looks weird and sticks out!). Such change would persist for all subsequent usages of same fingerprint spec.
 - SNI, however, does change regularly, and could start defaulting to a wrong thing, but only if not set explicitly with `SetSNI`.

This PR fixes statefulness by making utls generate a new `ClientHelloSpec` for each usage of default `ClientHelloSpec`s. Custom `ClientHelloSpec`s (which no one uses atm as far as I know), may still share *some* state, so I added a disclaimer for that to advise users to create a new custom ClientHelloSpec for each UConn.

### DIFF
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -209,8 +209,8 @@ func (uconn *UConn) applyPresetByID(id ClientHelloID) (err error) {
 }
 
 // ApplyPreset should only be used in conjunction with HelloCustom to apply custom specs.
-// Separate ClientHelloSpec for each UConn is advised, to avoid state sharing.
-// Also used internally.
+// Fields of TLSExtensions that are slices/pointers are shared across different connections with
+// same ClientHelloSpec. It is advised to use different specs and avoid any shared state.
 func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 	hello := uconn.HandshakeState.Hello
 	session := uconn.HandshakeState.Session
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -209,6 +209,7 @@ func (uconn *UConn) applyPresetByID(id ClientHelloID) (err error) {
 }
 
 // ApplyPreset should only be used in conjunction with HelloCustom to apply custom specs.
+// Separate ClientHelloSpec for each UConn is advised, to avoid state sharing.
 // Also used internally.
 func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 	hello := uconn.HandshakeState.Hello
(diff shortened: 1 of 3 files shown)


nikitaborisov:
This fixes the shared state problem I was seeing in gotapdance

u_parrots.go
```diff
@@ -208,6 +209,7 @@ func (uconn *UConn) applyPresetByID(id ClientHelloID) (err error) {
 }
 
 // ApplyPreset should only be used in conjunction with HelloCustom to apply custom specs.
+// Separate ClientHelloSpec for each UConn is advised, to avoid state sharing.
```
bemasc: Is this because `ApplyPreset` modifies `p`?  If so, that could be made explicit here.

u_parrots.go
```diff
@@ -250,15 +252,17 @@ func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 		uconn.greaseSeed[ssl_grease_extension2] ^= 0x1010
 	}
 
-	hello.CipherSuites = p.CipherSuites
+	hello.CipherSuites = make([]uint16, len(p.CipherSuites))
```
bemasc: This is confusing.  Are you deep-copying the input (in which case it's safe to use the same input for many calls) or are you requiring the input to be independent for each call (in which case there's no need to make a copy).

sergeyfrolov: The only reliable way to deepcopy all the extensions I found is to add `Clone()` method to `TLSExtension` interface. Which I didn't quite want to do.
Current solution simply makes a new preset for built-in parrots. For custom ones, only shared state is pointers and slices in fields of `TLSExtension`s, which I indeed could be more explicit about.

u_parrots.go
sergeyfrolov: It doesn't modify state of `p` anymore (at this time).

PR #9, opened 2018-07-30
Add Roller

sergeyfrolov:
NewRoller creates Roller object with default range of HelloIDs to cycle through until a
working/unblocked one is found.

Dial attempts to establish connection to given address using different HelloIDs.
If a working HelloID is found, it is used again for subsequent Dials.
If tcp connection fails or all HelloIDs are tried, returns with last error.
Usage examples:
   Dial("tcp4", "google.com:443", "google.com")
   Dial("tcp", "10.23.144.22:443", "mywebserver.org")


u_roller.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// NewRoller creates Roller object with default range of HelloIDs to cycle through until a
+// working/unblocked one is found.
+func NewRoller() (*Roller, error) {
+	tcpDialTimeoutInc, err := getRandUint16()
+	if err != nil {
+		return nil, err
+	}
+	tcpDialTimeoutInc = tcpDialTimeoutInc % 14
```
bemasc: I think this would be clearer as
```go
tcpDialTimeout := rand % 14 + 7
```
and similarly for other random constants.

u_roller.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, err
+	}
+	tlsHandshakeTimeoutInc = tlsHandshakeTimeoutInc % 20
+
+	r := &Roller{
+		HelloIDs: []ClientHelloID{
+			HelloChrome_Auto,
+			HelloFirefox_Auto,
+			HelloIOS_Auto,
+		},
```
bemasc: What about random?  I think it might make sense to include the randomized handshake as the last option, since that's the only one that "guaranteed" to work.

u_roller.go
```diff
@@ -0,0 +1,110 @@
...
+		TcpDialTimeout:      time.Second * time.Duration(7+int(tcpDialTimeoutInc)),
+		TlsHandshakeTimeout: time.Second * time.Duration(11+int(tlsHandshakeTimeoutInc)),
+	}
+
+	// choose random clientHelloID to try next
```
bemasc: Can you just shuffle `HelloIDs` instead?  That would be clearer.

u_roller.go
```diff
@@ -0,0 +1,110 @@
...
+			continue
+		}
+		c.HelloIDMu.Lock()
+		c.NextHelloID = nextHelloId
+		c.HelloIDMu.Unlock()
+		return client, err
+	}
+}
+
+func getRandUint16() (uint16, error) {
```
bemasc: It seems like all the calls to this function would be better off using [`rand.Intn`](https://golang.org/pkg/math/rand/#Rand.Intn) or [`crypto.rand.Int`](https://golang.org/pkg/crypto/rand/#Int)

u_roller.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil, err
+		}
+
+		client := UClient(tcpConn, nil, nextHelloId)
+		client.SetSNI(serverName)
+		client.SetDeadline(time.Now().Add(c.TlsHandshakeTimeout))
+		err = client.Handshake()
+		client.SetDeadline(time.Time{}) // unset timeout
+		if err != nil {
+			for i, ID := range remainingHelloIds {
```
bemasc: I can't follow this logic, but I think it's overcomplicated.  If you shuffle the array when it's created, then you can just walk down the array in order here.

u_roller.go
```diff
@@ -0,0 +1,110 @@
...
+	"encoding/binary"
+	"net"
+	"sync"
+	"time"
+)
+
+type Roller struct {
+	HelloIDs            []ClientHelloID
+	HelloIDMu           sync.Mutex
+	NextHelloID         ClientHelloID
```
bemasc: I think this would be easier to understand if instead of `NextHelloID` you had `WorkingHelloID`, i.e. a known-working HelloID that is initially `nil`.  Then there are two clear states: either you have a known working ID (use it) or you don't (go down the list).  Transitions between states are also pretty clear.

u_roller.go
sergeyfrolov: Good point.

u_roller.go
sergeyfrolov: Good point.

u_roller.go
sergeyfrolov: Not sure I agree. Instead of choosing random index, if I randomize the array I'd be able to always pick 0th element in it -- sounds like ~same level of clarity. Logic for "finding which element to evict from array" will have to remain, since last successful ID could be anywhere in shuffled array.

u_roller.go
sergeyfrolov: `rand.Intn` is insecure
`crypto.rand.Int` takes `rand.Reader`, and `big.Int` and returns `big.Int`(!) and `error`, which is a pain to deal with.
I can't see how either of those is better.

u_roller.go
sergeyfrolov: Not quite: I start with `NextHelloID`, such that we don't change IDs, when not necessary and stick to a working one while it works.

u_roller.go
bemasc: I don't know what the intended behavior of the roller is, because you haven't documented it.  However, I don't think the exact behavior is important, so it makes sense to choose the behavior that is clearest to implement.

I don't think there's any need to "evict elements from an array".  Instead, you could just check
1. Do I have a known working hello ID?  If so, use it.
2. Otherwise, try each entry in the list.
3. If one worked, record it as a known working ID.

If you don't like that heuristic, then there are plenty of other simple options like "If it worked, move it to the front of the list.".  The current heuristic seems unnecessarily complex.

u_roller.go
bemasc: If you don't like either of those, then I think you should implement an `Intn` style function instead of this one, and include a comment explaining why you didn't use either of the stdlib's options.

sergeyfrolov: Just remembered, I already made `getRandInt`, which takes simple `int` as maxValue, uses `big.Int` inside, and returns (`int`, `error`). This is a nice to use function, so I'll switch to that :)

u_roller.go
sergeyfrolov: Done

PR #14, opened 2019-01-03
 Implement TLS 1.3 Randomized Spec and revise 1.2 Randomized Spec

### LINKED ISSUE #13: "uTLS does not support 0x0 as min version" with `tls.HelloRandomized`
I'm using utls commit b84d7d5f055f78c6f3dc5cd75e459bb78f84cc83. Running this program:
```
package main

import (
	"fmt"
	"os"
	"net"

	tls "github.com/refraction-networking/utls"
)

func run() error {
	conn, err := net.Dial("tcp", "tlsfingerprint.io:443")
	if err != nil {
		return err
	}
	uconn := tls.UClient(conn, nil, tls.HelloRandomized)
	uconn.SetSNI("tlsfingerprint.io")
	return uconn.Handshake()
}

func main() {
	err := run()
	if err != nil {
		fmt.Fprintln(os.Stderr, err)
		os.Exit(1)
	}
}
```
Fails with the error message:
```
uTLS does not support 0x0 as min version
```
It also happens with `HelloRandomizedALPN` and `HelloRandomizedNoALPN`. It doesn't happen with `HelloChrome_70` and `HelloFirefox_63`.

sergeyfrolov:
Fixes #13


u_parrots.go
```diff
@@ -495,6 +495,25 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
 	if err != nil {
 		return p, err
 	}
+
+	if tossBiasedCoin(0.4) {
+		p.TLSVersMin = VersionTLS10
+		p.TLSVersMax = VersionTLS13
```
ewust: Is VersionTLS13 == VersionTLS12?

sergeyfrolov: no

u_parrots.go
```diff
@@ -567,22 +585,44 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
 		p.Extensions = append(p.Extensions, &alpn)
 	}
 
-	if tossBiasedCoin(0.66) {
+	if tossBiasedCoin(0.62) || p.TLSVersMax == VersionTLS13 {
+		// always include for TLS 1.3, since TLS 1.3 ClientHellos are often over 256 bytes
+		// and that's when padding is required to work around buggy middleboxes
```
ewust: But might we generate a client hello less than 256 bytes before padding? Then we might stick out?

u_parrots.go
```diff
@@ -567,22 +585,44 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
 		p.Extensions = append(p.Extensions, &padding)
 	}
-	if tossBiasedCoin(0.66) {
+	if tossBiasedCoin(0.74) {
```
ewust: I wouldn't worry about these weightings too much. It's not like exactly matching the underlying distribution is going to make the job any harder for the censor than if you just picked it with even odds.

u_parrots.go
```diff
@@ -624,7 +664,23 @@ func removeRandomCiphers(s []uint16, maxRemovalProbability float32) []uint16 {
...
+}
+
+func removeRC4Ciphers(s []uint16) []uint16 {
+	// removes elements in place
+	sliceLen := len(s)
+	for i := 0; i < sliceLen; i++ {
+		cipher := s[i]
+		if cipher == TLS_ECDHE_ECDSA_WITH_RC4_128_SHA ||
+			cipher == TLS_ECDHE_RSA_WITH_RC4_128_SHA ||
+			cipher == TLS_RSA_WITH_RC4_128_SHA {
```
ewust: Aren't there other RC4 ciphers? e.g. https://tlsfingerprint.io/find/cipher/c002
Might want to at least comment this assumes only a subset of all RC4 ciphers (there's a bunch of weak ones that presumably wouldn't ever get to this function, but it is named removeRC4Ciphers, not removeLikelyRC4: https://tlsfingerprint.io/weak-ciphers )

u_parrots.go
ewust: What version ends up in the handshake version field then? It seems like you'll be sending 0x0304 but you want to send 0x0303, or is that logic somewhere else?

u_parrots.go
sergeyfrolov: padding is a conditional extension that doesn't get added, if client hello size is not in the F5-terminator-failure-range

u_parrots.go
sergeyfrolov: i am not worrying too much, i am worrying a little :)

u_parrots.go
sergeyfrolov: those are the only RC4 ciphers implemented by Golang and returned by `defaultCipherSuites()`

u_parrots.go
ewust: So this block says: with 62% chance, add padding iff client hello size is in the F5-failure range? Or is it just saying "with 62% add padding"

u_parrots.go
sergeyfrolov: that logic is somewhere else, those fields are to store real versions and generate all the things accordingly

u_parrots.go
sergeyfrolov: > with 62% chance, add padding iff client hello size is in the F5-failure range

yes. if it's tls 1.2

if it's tls 1.3:
with 100% chance, add padding iff client hello size is in the F5-failure range

Issue #15, opened 2019-01-11
Randomized fingerprints, but consistent?


AxbB36:
`HelloRandomized` and friends provide a randomized fingerprint, but it is re-randomized for each new connection. Is there a way to get a "handle" to a specific randomized fingerprint so it can be re-used within a process?

Here's a sample program that makes 10 connections with `HelloRandomizedNoALPN`:
```
package main

import (
	"fmt"
	tls "github.com/refraction-networking/utls"
	"net"
)

const hostName = "golang.org"

func connect() error {
	conn, err := net.Dial("tcp", net.JoinHostPort(hostName, "443"))
	if err != nil {
		return err
	}
	uconn := tls.UClient(conn, nil, tls.HelloRandomizedNoALPN)
	uconn.SetSNI(hostName)
	err = uconn.Handshake()
	if err != nil {
		return err
	}
	return uconn.Close()
}

func main() {
	for i := 0; i < 10; i++ {
		err := connect()
		if err == nil {
			fmt.Println(i, "ok")
		} else {
			fmt.Println(i, err)
		}
	}
}
```
Sample output is:
```
0 ok
1 remote error: tls: handshake failure
2 ok
3 remote error: tls: handshake failure
4 ok
5 ok
6 ok
7 ok
8 ok
9 ok
```
I am using `tshark` to verify that the fingerprint changes with each connection:
```
tshark -V -f 'host golang.org' -Y ssl.handshake.ciphersuites
```

Incidentally, the reason this came up is because I'm trying to use `UClient` inside the `DialTLS` of [`http.Transport`](https://golang.org/pkg/net/http/#Transport) in order to get TLS fingerprint obfuscation along with connection reuse, keepalive, and the other features that `http.Transport` provides. When used with `utls`, `http.Transport` seems to create 10 independent connections (each with its own randomized fingerprint). I'll file a separate issue about that.

AxbB36:
> When used with `utls`, `http.Transport` seems to create 10 independent connections (each with its own randomized fingerprint). I'll file a separate issue about that.

Never mind on this minor point. If I read and close the `Body` of each response, then `http.Transport` uses only one connection, even when HTTP/1.1 is negotiated. (When HTTP/2 is negotiated, it uses only one connection, whether the bodies are read and closed or not.)

sergeyfrolov:
Implemented in #20

PR #20, opened 2019-02-10
Implement consistent randomized fingerprint

sergeyfrolov:
Implements #15 

After this change, ClientHelloID for randomized fp will hold a seed, and can be reused like this:
```Golang

newUtlsConn := tls.UClient(tcpConn, &config, oldUtlsConn.ClientHelloID) // reuses oldUtlsConn fingerprint ID
```

@AxbB36 @rod-hynes


AxbB36:
Thanks for doing this.

Looking at [the example](https://github.com/refraction-networking/utls/pull/20/files#diff-1f008566bc39eeadf765a2d47f5836e4 the example), it looks like the API is:
- Make a TCP connection.
- Call `UClient` with `HelloRandomized`/`HelloRandomizedALPN`/`HelloRandomizedNoALPN` and save the result.
- The call to `UClient` implicitly generates a randomized `ClientHelloID`.
- You can then make a second connection with the same fingerprint as the first, by calling `UClient` with the `ClientHelloID` generated by the first connection, in place of `HelloRandomized`/`HelloRandomizedALPN`/`HelloRandomizedNoALPN`.

It's surprising to me that you need to make a connection in order to generate a randomized fingerprint as a side effect. I think most real code is unlike the example in that there isn't a clear idea of which call to `UClient` is "first": instead, it will be called in a loop, or after a timeout. So in practice every call will have to have a check (actually much like [the check](https://github.com/refraction-networking/utls/pull/20/files#diff-17a6357df708d056c23f2eb7f0c63069R481) in `generateRandomizedSpec`) "Is this the first call? If so, use `HelloRandomized`. Otherwise, use the previous `ClientHelloID`." I was expecting something more like a factory function that I can call and it will produce a randomized `ClientHelloID` without making a connection, something like `generateRandomizedSpec` but as a public function. With the API as it is, I would be tempted to make a dummy localhost connection and throw it away, just for the sake of getting a randomized fingerprint that doesn't require case-checking later on.

What do you think about allowing the caller to supply their own `io.Reader` or `math/rand.Source` for random bits, instead of hardcoding `crypto/rand` and `math/rand`? That would be another way to provide determinism: the caller could for example call `math/rand.NewSource` with their own seed (which may even be cached in a file between executions, etc.). Or provide any other interface that satisfies the `math/rand.Source` interface, if they don't like the `math/rand` PRNG.

I am picturing something like this:
```
clientHelloID := tls.RandomizedID(false /*alpn*/, mrand.NewSource(1234))
```
Instead of taking separate options, it could take a struct with fields, to allow for expansion in the future.

sergeyfrolov:
> `clientHelloID := tls.RandomizedID(false /*alpn*/, mrand.NewSource(1234))`

I probably should have done a better job documenting this, but you can do something like this with current proposal.

You seem to know this, but just to make things explicit: `ClientHelloID` has 2 fields: `String` "Browser" and `int64` "Version". When you use "Randomized Browser", its version will be used as a seed. If randomized fingerprint version is 0 (default), it will generate a random version, use that version as a seed, and then store the whole generated ClientHelloID as uconn.ClientHelloID so you can reuse it.

So, as it stands now, you already could make `ClientHelloID` with pre-chosen "Version" and it will give you a randomized and consistent fingerprint. There's a problem both with this API and the API you pictured: user needs to actually try to establish a connection to determine that the generated fingerprint satisfies some requirements and works with a given server. Which is why I thought that dialing until a working/suitable fingerprint is found and then starting to reuse it would be an easy way to use Randomized fingerprint.

Anyhow, is there an advantage to use a `rand.Source` to initialize fingerprint, instead of the specifying a random seed in the Version?

P.S.
> "Is this the first call? If so, use `HelloRandomized`. Otherwise, use the previous `ClientHelloID`."

I am sitting on a patch to introduce this behavior to `utls.Roller`, will have more time after USENIX deadline this Friday.

AxbB36:
> > `clientHelloID := tls.RandomizedID(false /*alpn*/, mrand.NewSource(1234))`
> 
> I probably should have done a better job documenting this, but you can do something like this with current proposal.
> 
> You seem to know this, but just to make things explicit: `ClientHelloID` has 2 fields: `String` "Browser" and `int64` "Version". When you use "Randomized Browser", its version will be used as a seed. If randomized fingerprint version is 0 (default), it will generate a random version, use that version as a seed, and then store the whole generated ClientHelloID as uconn.ClientHelloID so you can reuse it.

Oh, actually that wasn't clear to me. I had been treating `ClientHelloID` as an opaque data type and didn't think to look at what public fields it had.

> So, as it stands now, you already could make `ClientHelloID` with pre-chosen "Version" and it will give you a randomized and consistent fingerprint. There's a problem both with this API and the API you pictured: user needs to actually try to establish a connection to determine that the generated fingerprint satisfies some requirements and works with a given server. Which is why I thought that dialing until a working/suitable fingerprint is found and then starting to reuse it would be an easy way to use Randomized fingerprint.

Okay. That's a good point about the randomized fingerprint possibly not working. I think it's fine if uTLS provides an automatic redialer, as long as the option to control a seed manually with no automatic behavior remains.

> Anyhow, is there an advantage to use a `rand.Source` to initialize fingerprint, instead of the specifying a random seed in the Version?

I'm just thinking of how the Psiphon-Labs fork uses their own custom PRNG: https://github.com/Psiphon-Labs/utls/commit/11a4cc03332235a55c37ed7d686f4ac008749ddb
I don't think that `mrand.Source` makes any claims about distinguishability--consider an adversary watching a sequence of fingerprints--so it would be nice to be able to drop in a CSPRNG and remove that possible distinguisher.

rod-hynes:
> I'm just thinking of how the Psiphon-Labs fork uses their own custom PRNG: [Psiphon-Labs@11a4cc0](https://github.com/Psiphon-Labs/utls/commit/11a4cc03332235a55c37ed7d686f4ac008749ddb)

Yes, we implemented repeatable randomized TLS fingerprints by passing in an optional seed that's used with our CSPRNG (chacha20-based).

Given the concerns mentioned in this PR about `math/rand`'s PRNG and distinguishability and the entropy of a 64-bit seed, we would consider retaining that customization in our fork.

> supply their own [...] `math/rand.Source`

It would be ideal if we could pass in our own `math/rand.Source`. But short of that, if utls were to use a stronger CSPRNG (one possibility is https://gitlab.com/yawning/obfs4/blob/master/common/drbg/hash_drbg.go, which implements `math/rand.Source`?) and seed, then I see no reason not to use stock utls instead of customizing.

Other Psiphon code takes responsibility for finding a "working" config, so we would not use a retrying dialer in utls. The reason for this is that we have a larger scope for the try-random-things/remember-what-works, beyond the TLS ClientHello. As long as we can initialize the seed for a new `UClient`, this will suffice for us.

Yawning:
> one possibility is `https://gitlab.com/yawning/obfs4/blob/master/common/drbg/hash_drbg.go`, which implements `math/rand.Source`?

It does implement that interface, yes.  That said, I would recommend against using it, as there are far better constructs that could be used, depending on what sort of properties are required.

sergeyfrolov:
Alright, I implemented SHA256 HKDF-based RNG for the randomized fingerprints in 79a37cfb6e8de35ea0df006ee8ec45b673db6319  
Everyone is welcome to provide an opinion, but I am especially interested to know if @rod-hynes considers this RNG strong enough to start using stock utls

Now the 63 bits of the seed are used (as opposed to previously used rand.NewSource() function, which takes int64 and then rounds it down to int32 throwing away precious entropy). I am not really concerned about entropy of 63 bit seed, tbh. For one, censor couldn't store so many fingerprints: even if every fp took only one byte (it takes way more than that, and if some of those fingerprints are the same, you'd have to store info about which ones are the same, which is still more than 1 byte), it'd take
```Golang
(2^63) bytes = 9223372.04 terabytes
```

u_parrots.go
```diff
@@ -619,38 +633,86 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
 	err = uconn.SetTLSVers(p.TLSVersMin, p.TLSVersMax)
 	if err != nil {
 		return p, err
 	}
 
-	return p, nil
+	return p, r.savedErr
+}
+
+// utlsRandomizer never returns errors, but stores first one in savedErr
```
Yawning: Why doesn't this implement `math/rand/Source` (Or `Source64`?).  The runtime library's `math/rand/Rand` provides perfectly good implementations of most of the helpers given a `Source`, except for the biased coin flip, that's trivial to implement given `Rand.Float32()`.

u_parrots.go
```diff
@@ -619,38 +633,86 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
+type utlsRandomizer struct {
+	r        io.Reader
+	savedErr error
+}
+
+func makeUtlsRandomizer(seed uint64) *utlsRandomizer {
+	seedBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(seedBytes, seed)
+	tdHkdf := hkdf.New(sha256.New, seedBytes, []byte("uTLS salty salt"), nil)
```
Yawning: Are all uses of any `utlsRandomizer` guaranteed to read <= 8160 bytes from the entropy source, ever?

u_parrots.go
```diff
@@ -619,38 +633,86 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
+	uintBytes := make([]byte, 8)
+	_, err := r.r.Read(uintBytes)
+	if r.savedErr == nil && err != nil {
+		r.savedErr = err
+	}
+	return binary.BigEndian.Uint64(uintBytes)
+}
+
+// returns int64 in [0, max)
+func (r *utlsRandomizer) randInt64n(max int) int64 {
```
Yawning: Should use the runtime library for this.

Output is non-uniform for most values of `n`.

u_parrots.go
```diff
@@ -619,38 +633,86 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
+	}
+	randUint := r.randUInt64() % uint64(max)
+	if randUint > math.MaxInt64 {
+		randUint >>= 1
+	}
+	return int64(randUint)
+}
+
+// returns int in [0, max) up to max int32
+func (r *utlsRandomizer) randIntn(max int) int {
```
Yawning: Use the runtime libray's implementation.

Output is non-uniform for most values of `n`.

u_parrots.go
```diff
@@ -619,38 +633,86 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
+		// if we ever do, it may panic on 32 bit architecture, so we have to check for that
+		r.savedErr = fmt.Errorf("utlsRandomizer does not support randIntn(%v): max is too big for a %v-bit architecture",
+			max, strconv.IntSize)
+		return 0
+	}
+	return int(r.randInt64n(max))
 }
 
-func tossBiasedCoin(probability float32) bool {
+func (r *utlsRandomizer) tossBiasedCoin(probability float32) bool {
```
Yawning: Implement on top of a `math/rand/Rand.Float32`.

u_parrots.go
```diff
@@ -619,38 +633,86 @@ func (uconn *UConn) generateRandomizedSpec(WithALPN bool) (ClientHelloSpec, erro
...
+	value := r.randIntn(precision)
 	if float32(value) <= threshold {
 		return true
 	} else {
 		return false
 	}
 }
 
-func removeRandomCiphers(s []uint16, maxRemovalProbability float32) []uint16 {
+func (r *utlsRandomizer) getRandPerm(n int) []int {
```
Yawning: Use the runtime library's implementation.

Output is currently biased due to `r.randIntn` being broken.

rod-hynes:
As Yawning pointed out, there are some deficiencies in this implementation including using modulus for `Intn`. Decent explanation here: https://stackoverflow.com/a/6852396.

You could fix this, but I strongly echo the recommendation to implement your CSPRNG as a `rand.Source64` (https://golang.org/pkg/math/rand/#Source64) and then create a `rand.Rand` using that source to get correct implementations of all the helpers.

The issue with the max number of bytes that may be read from HKDF concerns me. We use a chacha20 stream cipher.

As for forking, we're undecided. If you use `rand.Source64` internally, then we could make a very simple change in a fork in order to keep using our CSPRNG.  This has one benefit for us, in that existing Psiphon client that have persisted a seed for a working random TLS fingerprint may continue to generate the same fingerprint after we upgrade `utls` (for some cases, and depending on whether the order of calls to `rand.Rand` remain the same as before). If we go this route we could contribute a code change that enables this customization.

I'm not clear on the seed size argument.

> since if your image space [...] exceeds 2^64 by a large amount, it's trivial for a censor to tell if something is one of the 2^64 uTLS-generated "randomized" fingerprints vs not

What's the trivial attack? I assume it's not storing all fingerprints, as you pointed out?

In any case, we have a 256-bit seed at present and so downgrading that vs. minor fork customization is another consideration.

For the record, there remains a number of other small, unrelated customizations we probably need and will maintain a fork for. Ultimately I'd like to see everything upstream and use stock `utls`.

ewust:
The "trivial" attack for a censor to tell if something is one of the 2^64 fingerprints is not to store all the fingerprints, but to use a bloom filter to test if a given fingerprint is in the larger set. The censor could also maintain subsets of the 2^64 fingerprints and match probabilisticly.  Still an impractically large size for naive implementations, but I'll note that no sane cryptographic system uses 64-bits of security. If there's an easy way we could generate more than 2^64, we should. Otherwise, it's likely easier to block in other ways (e.g. keep state per client and observe multiple non-standard fingerprints).

u_common.go
```diff
@@ -167,3 +172,21 @@ func EnableWeakCiphers() {
 			suiteECDHE | suiteTLS12 | suiteDefaultOff | suiteSHA384, cipherAES, utlsMacSHA384, nil},
 	}...)
 }
+
+func getRandInt(max int) (int, error) {
+	bigInt, err := rand.Int(rand.Reader, big.NewInt(int64(max)))
+	return int(bigInt.Int64()), err
+}
+
+func getRandPerm(n int) ([]int, error) {
```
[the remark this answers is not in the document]
sergeyfrolov: Both of those are called in Roller

u_parrots.go
sergeyfrolov: I don't see why it would need to implement any interfaces, it's not even exposed to user.
Tbh, I am not so in love with math/rand after I've seen `NewSource()` round down int64 seed to int32 internally, meaning that I have to verify every function of theirs. Kinda seemed easier to just make a few short functions myself, but it is true that some of those implementations are perfectly good

sergeyfrolov:
@rod-hynes fair points
Would you like to submit Psiphon's implementation (which is currently GPL) of chacha-based prng replacing current hkdf prng as a pull request to utls's `consistRandom` branch?

> Ultimately I'd like to see everything upstream and use stock utls.

So do I! Please send PRs :)

rod-hynes:
> submit Psiphon's implementation (which is currently GPL) of chacha-based prng replacing current hkdf prng

Here is what that looks like: https://github.com/refraction-networking/utls/pull/24

Is this Google CLA still required: https://github.com/refraction-networking/utls/blob/master/CONTRIBUTING.md?

As suggested here, https://github.com/refraction-networking/utls/pull/20#discussion_r258316476, I added a new `Seed` field to `ClientHelloID` accommodate the larger seed. The "Must not be modified once set" comment refers to the fact that copies are made of `ClientHelloID`, which will copy the seed pointer but not make a copy of the array.

In the Psiphon case, I expect that we'll explicitly initialize `ClientHelloID.Seed`. The seed field and type is exported.

I pared down our `prng` package and changed the license for this minimized copy. I removed the global instance and most/all unused functions and made some types unexported.

Random helpers -- Intn, Perm, etc. -- now use the standard library.

sergeyfrolov:
@rod-hynes thanks for the contribution!
I merged a fix that was pushed to master in the meantime, and made some refactors in 21d729e0329c579a2a12a0983e028456639a992b: removed some redundant shuffle functions, and changed ClientHelloID.Version to be a string to incorporate versions like "12.3.10", and some other minor fixes.

Overall, this PR LGTM

PR #25, opened 2019-03-12
Update fingerprints + add default spec version

sergeyfrolov:
I am a bit too lazy to split those features, as they are tied together in the code.
 * Adds fingerprints for Chrome 75, iOS 12.1, and Firefox 65(=Firefox 63)
 * If min/max tls versions are not explicitly specified in the ClientHelloSpec,
   uTLS will try to parse versions from SupportedVersions extension,
   and fallback to [TLS 1.0, TLS 1.2] if SupportedVersions is absent.
 * Adds mimicked FakeRecordSizeLimitExtension and FakeCertCompressionAlgsExtension
   to be used instead of GenericExtension{} for clarity and extensibility
   (we are ready to use those with Firefox and Chrome fps with correct values
   whenever actual functionality is implemented)

### DIFF
--- a/u_conn.go
+++ b/u_conn.go
@@ -480,17 +480,21 @@ func (uconn *UConn) GetOutKeystream(length int) ([]byte, error) {
 	return nil, errors.New("Could not convert OutCipher to cipher.AEAD")
 }
 
-// SetVersCreateState set min and max TLS version in all appropriate places.
-// If minTLSVers == 0 && minTLSVers == 0,
-// SetTLSVers() will try to parse to parse the version from extensions.
-func (uconn *UConn) SetTLSVers(minTLSVers, maxTLSVers uint16, extensions []TLSExtension) error {
+// SetTLSVers sets min and max TLS version in all appropriate places.
+// Function will use first non-zero version parsed in following order:
+//   1) Provided minTLSVers, maxTLSVers
+//   2) specExtensions may have SupportedVersionsExtension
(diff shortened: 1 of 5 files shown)


u_conn.go
```diff
@@ -481,7 +481,51 @@ func (uconn *UConn) GetOutKeystream(length int) ([]byte, error) {
...
 
 // SetVersCreateState set min and max TLS version in all appropriate places.
-func (uconn *UConn) SetTLSVers(minTLSVers, maxTLSVers uint16) error {
+// If minTLSVers == 0 && minTLSVers == 0,
+// SetTLSVers() will try to parse to parse the version from extensions.
+func (uconn *UConn) SetTLSVers(minTLSVers, maxTLSVers uint16, extensions []TLSExtension) error {
+	if minTLSVers == 0 && maxTLSVers == 0 {
+		// if version is not set explicitly in the ClientHelloSpec, check the SupportedVersions extension
+		supportedVersionsExtensionsPresent := 0
+		for _, e := range uconn.Extensions {
```
rod-hynes:
```suggestion
		for _, e := range extensions {
```
I believe `uconn.Extensions` should be `extensions` here. As is, Chrome 72 fails with the error "tls: server selected unsupported protocol version 304" when it should negotiate TLS 1.3. With the change, Chrome 72 works ok in our test suite. iOS 12.1 also appears to work ok in our test suite.

PR #28, opened 2019-03-19
WIP: fix TLS 1.2 session cache

### LINKED ISSUE #27: Session tickets from ClientSessionCache are not used
In testing, it appears that session states/tickets set using `SetSessionCache` are not added to ClientHellos. Individual session states/tickets explicitly set using `SetSessionState` _are_ added to ClientHellos, as long as `SetSessionState` is called before `BuildHandshakeState`.

I suspect the issue with the cache is that `Handshake` calls `BuildHandshakeState` before `clientHandshake` is called here:

https://github.com/refraction-networking/utls/blob/f8425e69f7d4eff4e1988c6c4ca1ca9a6e17a42f/u_conn.go#L182-L189

`clientHandshake` consults the session cache and attempts to use session tickets, but its call to `SetSessionState` comes too late:

https://github.com/refraction-networking/utls/blob/f8425e69f7d4eff4e1988c6c4ca1ca9a6e17a42f/u_conn.go#L312-L334

There's a call to `hello.marshal` after the session cache is checked, but it will just use the previously marshaled value stored in `hello.Raw` set in `MarshalClientHello` which is called by `BuildHandshakeState`:

https://github.com/refraction-networking/utls/blob/f8425e69f7d4eff4e1988c6c4ca1ca9a6e17a42f/u_conn.go#L468

As a result, it appears that TLS session resumption won't work, as session tickets received from servers and stored in the cache will never be sent in ClientHellos?

I didn't try setting a cache directly using `Config.ClientSessionCache`; it appears that would have the same issue?

sergeyfrolov:
Currently, SessionCache is not used for non-HelloGolang fingerprints,
and this PR is what the fix would potentially look like.
This is only for the TLS 1.2, support for TLS 1.3 PSK resumption cache remains a TODO.

Supposedly fixes #27, not tested yet

@rod-hynes

### DIFF
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -427,6 +427,11 @@ func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 			}
 			grease_extensions_seen += 1
 		case *SessionTicketExtension:
+			if session == nil && uconn.config.ClientSessionCache != nil {
+				cacheKey := clientSessionCacheKey(uconn.RemoteAddr(), uconn.config)
+				session, _ = uconn.config.ClientSessionCache.Get(cacheKey)
+				// TODO: use uconn.loadSession(hello.getPrivatePtr()) to support TLS 1.3 PSK-style resumption
+			}
 			err := uconn.SetSessionState(session)
 			if err != nil {
 				return err


rod-hynes:
This change appears to fix my test case in https://github.com/refraction-networking/utls/issues/27. Thanks.

PR #30, opened 2019-04-14
Fix panic in u_public by never dereferencing

### LINKED ISSUE #29: Panic in toPublic12
We have encountered this panic in `utls`:

> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd1690b88]
> 
> goroutine 24162 [running]:
> github.com/Psiphon-Labs/psiphon-tunnel-core/vendor/github.com/refraction- networking/utls.(*clientHandshakeState).toPublic12(0x935a81e0, 0xd1f18f78)
>      /go/src/github.com/Psiphon-Labs/psiphon-tunnel-core/vendor/github.com/refraction- networking/utls/u_public.go:136 +0x12c
> github.com/Psiphon-Labs/psiphon-tunnel-core/vendor/github.com/refraction-networking/utls.(*UConn).clientHandshake(0x934ea2d0, 0xd1f18f78, 0x936b40e0)
>      /go/src/github.com/Psiphon-Labs/psiphon-tunnel-core/vendor/github.com/refraction-networking/utls/u_conn.go:372 +0x4a4
> github.com/Psiphon-Labs/psiphon-tunnel-core/vendor/github.com/refraction-networking/utls.(*UConn).Handshake(0x934ea2d0, 0x0, 0x0)
>      /go/src/github.com/Psiphon-Labs/psiphon-tunnel-core/vendor/github.com/refraction-networking/utls/u_conn.go:189 +0x11c

The panic is here:
https://github.com/refraction-networking/utls/blob/b7c656eec2d2aa957b1e844446e867a567e17d1f/u_public.go#L131-L136

I suspect `suite` is nil?

`toPublic12` is called from here, and is called even when `hs12.handshake` returns an error:
https://github.com/refraction-networking/utls/blob/b7c656eec2d2aa957b1e844446e867a567e17d1f/u_conn.go#L371-L375

`handshake` calls `processServerHello` which  calls `pickCipherSuite` which can fail and leave `suite` nil:
https://github.com/refraction-networking/utls/blob/b7c656eec2d2aa957b1e844446e867a567e17d1f/handshake_client.go#L429-L433

So it appears that when the "server chose an unconfigured cipher suite" error occurs, `suite` is nil and `toPublic12` will panic since it's called before the `err` check?

There's a similar case where `toPublic13` is called before checking `err` here:
https://github.com/refraction-networking/utls/blob/b7c656eec2d2aa957b1e844446e867a567e17d1f/u_conn.go#L363-L365

sergeyfrolov:
In order to avoid panics, we make sure that no pointer is dereferenced
during transformation between public and private handshake state.
Presumably fixes #29

### DIFF
--- a/u_conn.go
+++ b/u_conn.go
@@ -361,7 +361,9 @@ func (c *UConn) clientHandshake() (err error) {
 		}
 		// In TLS 1.3, session tickets are delivered after the handshake.
 		err = hs13.handshake()
-		c.HandshakeState = *hs13.toPublic13()
+		if handshakeState := hs13.toPublic13(); handshakeState != nil {
+			c.HandshakeState = *handshakeState
+		}
 		return err
 	}
 
@@ -369,7 +371,9 @@ func (c *UConn) clientHandshake() (err error) {
 	hs12.serverHello = serverHello
 	hs12.hello = hello
 	err = hs12.handshake()
-	c.HandshakeState = *hs12.toPublic12()
+	if handshakeState := hs12.toPublic12(); handshakeState != nil {
+		c.HandshakeState = *handshakeState
+	}
 	if err != nil {
 		return err
 	}
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -567,7 +567,7 @@ func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 			if session == nil && uconn.config.ClientSessionCache != nil {
 				cacheKey := clientSessionCacheKey(uconn.RemoteAddr(), uconn.config)
 				session, _ = uconn.config.ClientSessionCache.Get(cacheKey)
-				// TODO: use uconn.loadSession(hello.getPrivatePtr()) to support TLS 1.3 PSK-style resumption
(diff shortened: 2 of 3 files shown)


rod-hynes:
I believe this change does fix https://github.com/refraction-networking/utls/issues/29.

I tested against a TLS server configured to select a "FAKE" cipher suite. Without this change, I reproduced the panic with the same stack. With this change, there's no panic and I receive the expected "tls: server chose an unconfigured cipher suite" error return value.

Out of curiosity: since `TLS13OnlyState.Suite` is a pointer, there's no nil dereference in `toPublic13` and the code in that case didn't have to change. Why not simply change `TLS12OnlyState.Suite/FinishedHandshake` to be pointers as well so that, like the TLS 1.3 case, there is no nil dereference and no need for object copies?

sergeyfrolov:
`finishedHash` in internal `clientHandshakeState` is an object, not a pointer, so turning `FinishedHash` to a pointer in public handshake state would just mean more checks. Also pointers inherently adds a bit of inconvenience to end user.
`suite` is a pointer in internal struct though, but not sure if turning it into a pointer in public handshake state would improve consistency, given the `finishedHash` situation.

PR #31, opened 2019-08-12
Fix all-zeroes SessionID

rod-hynes:


rod-hynes:
The above test failure still occurs, repeatedly, even if I revert my commit:
```
--- FAIL: TestVerifyHostnameResumed (1.05s)
    --- FAIL: TestVerifyHostnameResumed/TLSv12 (0.52s)
        tls_test.go:380: Subsequent connection unexpectedly didn't resume
    --- FAIL: TestVerifyHostnameResumed/TLSv13 (0.52s)
        tls_test.go:380: Subsequent connection unexpectedly didn't resume
```
It looks like there are external dependencies: https://github.com/refraction-networking/utls/blob/master/tls_test.go#L361.

I tried running the stock `crypto/tls` test suite (Go v1.11.12) and that same test case passes.

sergeyfrolov:
The test in question has been deemed unreliable and was removed in upstream https://go-review.googlesource.com/c/go/+/186239/ 
> crypto/tls: remove TestVerifyHostnameResumed

> Session resumption is not a reliable TLS behavior: the server can decide
> to reject a session ticket for a number of reasons, or no reason at all.
> This makes this non-hermetic test extremely brittle.

> It's currently broken on the builders for both TLS 1.2 and TLS 1.3

Eventually, we'll merge the removal of this test from upstream, together with other upstream changes.
For now, I'll just merge this PR.
Thanks!

Issue #36, opened 2019-09-03
Replace Yawning/chacha20 PRNG


renzhexigua:
In the latest commit: https://github.com/Yawning/chacha20/commit/644b09ac4e6e3ed1c52aa43dad519087d8bca406, the author did some non backward-compatible changes. One of them is:

> `NewCipher` -> `New`

We should fix this piece here:

https://github.com/refraction-networking/utls/blob/7c97cdb47655017af7126d6c5cd23ace4469d7aa/u_prng.go#L119-L120

Yawning:
So, as a fair warning, the commit in question also changes the license to something that's probably not suitable for this project as well.

That said, in the spirit of "sorry I broke your package":

 * Using go modules (since utls requires a new-ish version of the toolchain anyway) will prevent dependency changes from breaking code in the future in this manner (till the dependencies are updated), and would also let you pin the chacha20 import to the previous commit.

 * https://godoc.org/github.com/aead/chacha20 is probably a suitable replacement, if people want to stick to ChaCha20, though skimming the code it does panic on counter overflow.  Personally I opted to save an import and just use SHAKE.

renzhexigua:
A useful circumvention way.

sergeyfrolov:
So, uTLS package is broken.

I do not think pinning chacha20 import to a previous commit is a good idea going forward. We'll have to find a replacement.

>  the commit in question also changes the license to something that's probably not suitable for this project as well

@rod-hynes does this affect you in any way? Do you happen to know any good replacement PRNG packages?

Seems like SHAKE (https://godoc.org/golang.org/x/crypto/sha3) could indeed be a good replacement.

rod-hynes:
> @rod-hynes does this affect you in any way?

We don't have any immediate issues since we vendored an earlier version of this: https://github.com/Psiphon-Labs/psiphon-tunnel-core/tree/master/vendor/github.com/Yawning/chacha20.

SHAKE seems fine and should simplify the code in https://github.com/refraction-networking/utls/blob/master/u_prng.go since there's no stream limit, so no need to rekey. I think SHAKE is somewhat slower than chacha20 (https://bench.cr.yp.to), but that's not relevant here, with few bytes generated.

Changing the cipher means a one-time change in randomized fingerprints generated by persisted seeds. I think we're ok with that. So as long as the API remains the same -- can the same seed size be used? Something like:

```
p.randomStream := NewShake256()
_, err := p.randomStream.Write(seed[:])
```

[...]

PR #37, opened 2019-09-09
Use SHAKE instead of chacha20 for PRNG stream

rod-hynes:
See: https://github.com/refraction-networking/utls/issues/36

### DIFF
--- a/u_prng.go
+++ b/u_prng.go
@@ -14,11 +14,12 @@ package tls
 import (
 	crypto_rand "crypto/rand"
 	"encoding/binary"
+	"io"
 	"math"
 	"math/rand"
 	"sync"
 
-	"github.com/Yawning/chacha20"
+	"golang.org/x/crypto/sha3"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


rod-hynes:
Note: `newPRNGWithSeed` now returns an error, as I propagate the error return from `ShakeHash.Write` -- even though the current implementation in `sha3` never returns an error: https://github.com/golang/crypto/blob/81e90905daefcd6fd217b62423c0908922eadb30/sha3/sha3.go#L123.

The test failure is the same as before: https://github.com/refraction-networking/utls/pull/31#issuecomment-520595396.

u_prng.go
```diff
@@ -69,68 +61,36 @@ func newPRNG() (*prng, error) {
...
-	}
-
-	p.randomStream.KeyStream(b)
-
-	p.randomStreamUsed += uint64(len(b))
+	// ShakeHash.Read never returns an error:
+	// https://godoc.org/golang.org/x/crypto/sha3#ShakeHash
+	_, _ = io.ReadFull(p.randomStream, b)
 
 	return len(b), nil
```
sergeyfrolov: Why not return the error from `io.ReadFull()` anyway?
I don't think it is likely that Read() will start returning errors, but it is a package in `/x/` after all, meaning they can break things.

u_prng.go
```diff
@@ -69,68 +61,36 @@ func newPRNG() (*prng, error) {
...
 		return nil, err
 	}
-	return newPRNGWithSeed(seed), nil
+	return newPRNGWithSeed(seed)
 }
 
 // newPRNGWithSeed initializes a new PRNG using an existing seed.
-func newPRNGWithSeed(seed *PRNGSeed) *prng {
+func newPRNGWithSeed(seed *PRNGSeed) (*prng, error) {
+	shake := sha3.NewShake256()
```
sergeyfrolov: Would Shake128 be enough? I have no objections against 256, but some uTLS users seemed to be concerned about performance.

u_prng.go
rod-hynes: That would mean propagating that error back through all the PRNG helper functions, so every `FlipWeightedCoin` etc. call would have to check errors. One of the nice properties of the original implementation was that there are no error returns on the helpers.

I think the Go convention of the documentation being a contract means it's reasonably safe to assume that this won't be changed lightly:

"// Read... It never returns an error.": https://godoc.org/golang.org/x/crypto/sha3#ShakeHash

u_prng.go
rod-hynes: This comment about image space is probably relevant:

> if your image space (e.g. the permutations of extensions/cipher suites/etc) exceeds 2^64 by a large amount...

https://github.com/refraction-networking/utls/pull/20#discussion_r258316476

We have a 256-bit seed, so sticking with 256-bit security strength (https://godoc.org/golang.org/x/crypto/sha3#NewShake256) makes sense, I think.

u_prng.go
sergeyfrolov: > That would mean propagating that error back through all the PRNG helper functions, so every FlipWeightedCoin etc. call would have to check errors.

We can zero out (by setting it to `_`) the already-nil-error in those functions, instead of ignoring it inside of `(p *prng) Read()`. In other words, uTLS can offer the same contract: `prng.Read()` always returns the nil error, but it relies on the ShakeHash.Read() contract to do that. This way, if ShakeHash's contract is ever violated, prng will at least return an error, for whatever that is worth.
Feel free to leave it as it is.

sergeyfrolov:
Thanks for this contribution!

PR #38, opened 2019-10-01
Correctly return the SNI extension length when not using SNI

myleshorton:
This is a small change allowing deployments to not use SNI if they so choose (advantageous in some scenarios).


myleshorton:
Hmmm...seems to break some tests. Can't look atm unfortunately.

sergeyfrolov:
There is no need to explain that omitting SNI could be advantageous, I am happy to keep supporting omitted SNI, and potentially improve API for that. But what would be useful is to explain how you expect to use this, and what this change is supposed to do under which conditions. It appears you're trying to use one of the parrots, that includes SNI, but you don't set an SNI (or "set" an empty one, which is a no-op), and expect SNI to not be sent.

An ostensibly cleaner way to omit SNI is to remove the `SNIExtension` from the `uconn.Extensions` list before marshalling or not put it in there in the first place.

If there's an interest, we can discuss API improvements to make omitting SNI easier. We can design a function like `RemoveSNI()` to remove `SNIExtension`s from the list. There might be some confusion with SNI-that-is-sent and SNI-that-is-used-to-verify. For that reason, I'd probably prefer a `SetOmittedSNI(sni string)` function that would remove the extension from the list, but still set the SNI that will be used for verification, and then leave a comment for users to explain that they would have to explicitly set `InsecureSkipVerify` if they also choose to not verify the SNI.

myleshorton:
Hey sorry for the delay on this @sergeyfrolov but will play around with those options in our own deployment next week.

myleshorton:
Hey sorry looping back around on this finally @sergeyfrolov. I guess the nice part about the proposed change on the SNI length is that users don't have to do anything uTLS-specific for it to work. Simply omitting ServerName in the Config will do the right thing. 

Didn't quite realize that our other few changes would automatically track here, but we've found exposing a couple more things here and there to be useful.

PR #50, opened 2020-07-22
Fix openssl command arguments for update tests

max-b:
I've been trying to run the `-update` tests in order to fix the testdata for https://github.com/refraction-networking/utls/pull/38 and ran into a couple issues that I hope this PR addresses.

The first is that it seems as though the openssl command as populated currently ends up as:
```
openssl s_server -no_ticket -num_tickets 0 openssl s_server -cipher TLS_AES_128_GCM_SHA256 -tls1_3 -cert /tmp/go-tls-test334278555 -certform DER -key /tmp/go-tls-test012468798 -accept 24323
```
where `openssl` is repeated. It looks to me like the `openssl s_server` is already defined in https://github.com/refraction-networking/utls/blob/master/handshake_client_test.go#L170 so we can remove the `"openssl", "s_server"` values from the `args` slice.

The other is changing the `-cipher` flag to `-ciphersuites` for TLSv1.3. From https://www.openssl.org/docs/man1.1.1/man1/s_server.html 
```
-cipher val

    This allows the list of TLSv1.2 and below ciphersuites used by the server to be modified. This list is combined with any TLSv1.3 ciphersuites that have been configured. When the client sends a list of supported ciphers the first client cipher also included in the server list is used. Because the client specifies the preference order, the order of the server cipherlist is irrelevant. See the ciphers command for more information.
-ciphersuites val

    This allows the list of TLSv1.3 ciphersuites used by the server to be modified. This list is combined with any TLSv1.2 and below ciphersuites that have been configured. When the client sends a list of supported ciphers the first client cipher also included in the server list is used. Because the client specifies the preference order, the order of the server cipherlist is irrelevant. See the ciphers command for more information. The format for this list is a simple colon (":") separated list of TLSv1.3 ciphersuite names.
```

One note I should make is that running the tests with the `-update` flag changes the testdata from the current `master`. I suppose they would no matter what because the server random will change, so I haven't included those diffs in this PR. 

Let me know what you think/if there's anything else you'd like to see. Thanks!!

### DIFF
--- a/u_conn_test.go
+++ b/u_conn_test.go
@@ -135,7 +135,7 @@ func TestUTLSHandshakeClientParrotChrome_58_setclienthello(t *testing.T) {
 	opensslCipherName := "ECDHE-RSA-AES128-GCM-SHA256"
 	test := &clientTest{
 		name:    "UTLS-setclienthello-" + opensslCipherName + "-" + helloID.Str(),
-		args: []string{"openssl", "s_server", "-cipher", opensslCipherName},
+		args: []string{"-cipher", opensslCipherName},
 		config:  config,
 	}
 
@@ -152,7 +152,7 @@ func TestUTLSHelloRetryRequest(t *testing.T) {
 
 	test := &clientTest{
 		name:    "UTLS-HelloRetryRequest-" + helloID.Str(),
-		args: []string{"openssl", "s_server", "-cipher", "ECDHE-RSA-AES128-GCM-SHA256", "-curves", "P-256"},
+		args: []string{"-cipher", "ECDHE-RSA-AES128-GCM-SHA256", "-curves", "P-256"},
 		config:  config,
 	}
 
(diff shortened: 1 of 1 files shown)


u_conn_test.go
```diff
@@ -290,7 +290,7 @@ func testUTLSHandshakeClientECDHE_RSA_AES256_GCM_SHA256(t *testing.T, helloID Cl
 	opensslCipherName := "ECDHE-RSA-AES128-GCM-SHA256"
 	test := &clientTest{
 		name:    "UTLS-" + opensslCipherName + "-" + helloID.Str(),
-		args: []string{"openssl", "s_server", "-cipher", opensslCipherName},
+		args: []string{"-cipher", opensslCipherName},
```
sergeyfrolov, line 293: is there a reason to have `-cipher` here, but `-ciphersuites` below?

max-b: If I'm correctly understanding the docs, `-cipher` is for tlsv1.2 and below and `-ciphersuites` is for tlsv1.3.

max-b:
Thanks!

PR #51, opened 2020-07-22
Allow empty sni extension

max-b:
This attempts to get merged just the empty SNI changes from https://github.com/refraction-networking/utls/pull/38

I used the changes from https://github.com/refraction-networking/utls/pull/50 to generate the `testdata`

It looked to me like you had given the :+1: to the empty SNI changes in #38, but let me know if I misread that or if there's anything else you'd like incorporated here.

Thanks!


sergeyfrolov:
Thanks for the PR!

I did give :+1: to the empty SNI implementation in https://github.com/refraction-networking/utls/pull/38#issuecomment-537675514 , but in a different way, it seems. If I understand this correctly, the behavior of existing applications that use uTLS with empty SNIs will change with this PR, which is undesirable. We can implement omitting of SNI in a backward compatible way  by adding a new `RemoveSNIExtension()` function. As for the `testdata`, it would be useful to add a test for the`RemoveSNIExtension()` function and add a new test trace for it, but all other traces should not change.

max-b:
Great ok that makes sense - thanks for the feedback! I'll get on that.

max-b:
Ok so I reverted the existing commits on this branch (I can either rebase this branch or you can just use github's "squash and merge" functionality if you'd like.

It looked to me like it wouldn't work to have the `RemoveSNIExtension` work if the `ClientHelloID` was a `HelloGolang`, but it seems totally possible that I've missed something.

u_conn.go
```diff
@@ -162,6 +167,26 @@ func (uconn *UConn) SetSNI(sni string) {
...
 
+// RemoveSNIExtension will cause the UConn connection client hello to not include a SNI extension
+// This only applies to non-HelloGolang ClientHelloIDs
+func (uconn *UConn) RemoveSNIExtension() error {
+	if uconn.ClientHelloID == HelloGolang {
+		return fmt.Errorf("Cannot call RemoveSNIExtension on a UConn with a HelloGolang ClientHelloID")
+	}
+	uconn.omitSNIExtension = true
+	return nil
+}
```
max-b, line 178: This returns an `error` to prevent someone from accidentally trying to use it on a `UConn` that has a `HelloGolang` `ClientHelloID`

u_conn_test.go
```diff
@@ -365,14 +365,26 @@ func testUTLSHandshakeClientECDHE_ECDSA_WITH_CHACHA20_POLY1305(t *testing.T, hel
...
 func runUTLSClientTestForVersion(t *testing.T, template *clientTest, prefix, option string, helloID ClientHelloID) {
-	test := *template
-	test.name = prefix + test.name
-	if len(test.args) == 0 {
-		test.args = defaultClientCommand
+	run := func(omitSNI bool) {
+		test := *template
+		// clone the test template config so we can successfully re-run
+		templateConfig := *template.config // nolint
+		test.config = &templateConfig
```
max-b, line 372: Is this a reasonable way to clone the test config for re-running it?

max-b:
> It would be nicer if the test was in it's own TestRemoveSNIExtension function such that we could understand immediately if it broke. So, we should not clone the test config.
> 
> It also doesn't need to double the amount of test cases just for that feature, I think a single test (with say latest Chrome) is probably enough. My opinion on that is less strong on this one and I am happy to accept a PR that doubles test data, just does not seem necessary.

Works for me - I agree that's a better approach. Let me know if I've missed anything else.

Thanks!

max-b:
Great thanks so much!!

Issue #54, opened 2020-07-24
Use recorded ClientHello messages as templates/fingerprints


max-b:
We think it could be useful to generate client hello messages from recorded bytes.

Ideally we would like to contribute this upstream, so I wanted to see if you had any thoughts/concerns/ideas about this approach towards the outset.

We had two ideas about how we might proceed:

### Idea 1
Attempt to create a `ClientHelloSpec` from a `[]byte` slice, something along the lines of: `func FingerPrintClientHello(data []byte) (*ClientHelloSpec, error)`

This seems to mostly require figuring out how to populate the `[]TLSExtension` slice. We could use the `clientHelloMsg.unmarshal` to get information about the Extensions that go's `crypto/tls` library understands/parses, but that leaves us with at least a handful of `[]TLSExtension`s that we'd have to go back through the raw bytes and attempt to re-parse. As we re-parse the hello, we could look specifically for extensions which go doesn't support/parse and then use the `GenericExtension` type to just stuff those bytes into our hello.

It occurs to me that re-parsing likely suffers from 2 problems:
1. I _think_ that our extension could end up out of order from the recorded one?
2. We're basically going to have to move through the whole client hello byte slice anyways in order to parse the generic extensions

At this point it might be worth just writing a parallel `unmarshal` function which does what we need, but I guess I get a little nervy at making a parallel copy of so much core `crypto/tls` code?

### Idea 2
Use the recorded bytes to populate a `clientHelloMsg.raw`, unmarshal those bytes to fill out the rest of the relevant `clientHelloMsg` fields, update `Config` and `UConn` state to reflect any necessary changes, and update the `ClientHello` as appropriate (random, etc). 
In the [Custom Handshake section of the README](https://github.com/refraction-networking/utls#custom-handshake), it says:

> If you need to manually control all the bytes on the wire(certainly not recommended!), you can set UConn.HandshakeStateBuilt = true, and marshal clientHello into UConn.HandshakeState.Hello.raw yourself. In this case you will be responsible for modifying other parts of Config and ClientHelloMsg to reflect your setup and not confuse "crypto/tls", which will be processing response from server.

I'm not actually seeing that field on the `UConn` struct? Additionally, it looks to me that currently `BuildHandshakeState()` and then consequently the `MarshalClientHello()` are called by `Handshake()` and that they will modify raw bytes in any constructed `ClientHelloMsg`

[...]

PR #58, opened 2020-08-14
Implement FingerprintClientHello to generate ClientHelloSpec from ClientHello raw bytes

max-b:
My attempt to implement https://github.com/refraction-networking/utls/issues/54

We made a handful of assumptions/decisions and we're eager for your feedback.

The `FingerprintClientHello` as we've written it is somewhat conservative in that it returns an error if it encounters extensions it doesn't know how to properly mimic. That being said, it looks to properly handle 17 out of the top 20 fingerprints from https://tlsfingerprint.io/top/fingerprints

3 of those top 20 have a "generic"(unknown) extension that will currently fail our spec generation.
https://tlsfingerprint.io/id/5408690af1e08199
https://tlsfingerprint.io/id/247451152181eae6
https://tlsfingerprint.io/id/1fa228e941491e51

Those extensions are 0x0018, 0x000f, and 0x0029. I can look more closely into those extensions in particular to see if they have any accepted understanding. I think we'll also be doing some analysis of our clients to confirm that their client hello generation roughly matches the data you have.

We had also considered potentially adding a parameter for whether to allow/accept unknown extensions (or in general add a strategy for loosening restrictions) for users who would be less risk averse/more sure of their destination's handling.


sergeyfrolov:
> The FingerprintClientHello as we've written it is somewhat conservative in that it returns an error if it encounters extensions it doesn't know how to properly mimic.

Good idea.

> Those extensions are 0x0018, 0x000f, and 0x0029. I can look more closely into those extensions in particular to see if they have any accepted understanding. I think we'll also be doing some analysis of our clients to confirm that their client hello generation roughly matches the data you have.
> pre_shared_key (0x0029)

Seems like we can just remove it from the spec? Worth confirming, but I'd guess clients are probably only adding this extension to their regular fingerprint at a specific when they do have the PSK, but otherwise might look the same. Ofc, it's probably useful to check first if the addition of PSK affects other extensions. OTOH, I'd also guess that if it some device sends fingerprints with PSK, it probably will send the same fingerprint without it at some point.

> token_binding (0x0018)

This one is not implemented by crypto/tls nor uTLS, so makes sense to leave it for now

> heartbeat (0x000f)

Also not implemented, and I think popular implementations are staying away from it.
It's on a non-so-common fingerprint that looks like a scanner: https://tlsfingerprint.io/id/247451152181eae6

>  I think we'll also be doing some analysis of our clients to confirm that their client hello generation roughly matches the data you have.

That's a good idea, since I likely won't have time to do such analysis myself.

> We had also considered potentially adding a parameter for whether to allow/accept unknown extensions (or in general add a strategy for loosening restrictions) for users who would be less risk averse/more sure of their destination's handling.

Description of this feature deserves a big warning, but sounds useful!

Overall, it may make sense to implement this function on a struct so we can configure it, along the lines of
```Golang
type Fingerprinter struct {
  KeepPSK bool // such that the default is false
  AllowBluntMimicry bool
}

func (f* Fingerprinter) FingerprintClientHello()
```
and move it to a separate u_fingerprinter.go file

max-b:
Thanks for the feedback! I went ahead w/ your suggested changes. I have maybe one or two specific questions that I think I'll just highlight inline.

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+
+		case utlsExtensionPadding:
+			clientHelloSpec.Extensions = append(clientHelloSpec.Extensions, &UtlsPaddingExtension{GetPaddingLen: BoringPaddingStyle})
+
+		case fakeExtensionChannelID, fakeCertCompressionAlgs, fakeRecordSizeLimit:
+			clientHelloSpec.Extensions = append(clientHelloSpec.Extensions, &GenericExtension{extension, extData})
+
+		case extensionPreSharedKey:
+			// RFC 8446, Section 4.2.11
+			if f.KeepPSK || f.AllowBluntMimicry {
```
max-b, line 266: I wasn't sure whether this should be an `OR` of these conditions ie. whether `AllowBluntMimicry` should also pass along the preshared key data.

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
+// Copyright 2017 Google Inc. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
max-b, line 3: I assume this is the proper header?

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+
+	"golang.org/x/crypto/cryptobyte"
+)
+
+// Fingerprinter is a struct largely for holding options for the FingerprintClientHello func
+type Fingerprinter struct {
+	// KeepPSK will ensure that PSK extensions are passed along into the resulting ClientHelloSpec
+	KeepPSK bool
+	// AllowBluntMimicry will ensure that unknown extensions are passed along into the resulting ClientHelloSpec
+	// WARNING: there could be numerous subtle issues with ClientHelloSpecs that are generated with this flag
```
max-b, line 20: Sufficient warning??

u_fingerprinter_test.go
```diff
@@ -0,0 +1,408 @@
...
+			if genericExtension.Id == extensionId && bytes.Equal(genericExtension.Data, extensionData) {
+				return
+			}
+		}
+	}
+	t.Errorf("generated ClientHelloSpec with BluntMimicry did not correctly carry over generic extension")
+}
+
+func TestUTLSFingerprintClientHelloKeepPSK(t *testing.T) {
+	// Handshake Protocol: Client Hello
```
max-b, line 264: I did this in some `u_conn.go` tests as well. Do you think this is a reasonable format for using a raw/captured Client Hello?

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+	AllowBluntMimicry bool
+}
+
+// FingerprintClientHello returns a ClientHelloSpec which is based on the ClientHello that is passed in as the data argument
+//
+// If the ClientHello passed in has extensions that are not recognized or cannot be handled, it will return a non-nil error and a nil *ClientHelloSpec value
+//
+// The data should be the ClientHello record as produced by crypto/tls.clientHelloMsg.marshal()
+// ie. it should not contain the full tls record, just the handshake message as outlined in https://tools.ietf.org/html/rfc5246#section-7.4
+func (f *Fingerprinter) FingerprintClientHello(data []byte) (*ClientHelloSpec, error) {
```
sergeyfrolov, line 30: Have you parsed the minTLSVers, maxTLSVers, which are part of ClientHelloSpec?
It's possible that the code is doing the right thing, given the behavior of [SetTLSVers()](https://github.com/refraction-networking/utls/blob/186025ac7b77465439618d1aeb2a5e444714d1cc/u_conn.go#L512), just wanted to double check.

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+	// KeepPSK will ensure that PSK extensions are passed along into the resulting ClientHelloSpec
+	KeepPSK bool
+	// AllowBluntMimicry will ensure that unknown extensions are passed along into the resulting ClientHelloSpec
+	// WARNING: there could be numerous subtle issues with ClientHelloSpecs that are generated with this flag
+	AllowBluntMimicry bool
+}
+
+// FingerprintClientHello returns a ClientHelloSpec which is based on the ClientHello that is passed in as the data argument
+//
+// If the ClientHello passed in has extensions that are not recognized or cannot be handled, it will return a non-nil error and a nil *ClientHelloSpec value
```
sergeyfrolov, line 26: Wrap the comment line, here and elsewhere

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+//
+// The data should be the ClientHello record as produced by crypto/tls.clientHelloMsg.marshal()
```
sergeyfrolov, line 28: why not the full record?
that should make the function easier to use

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+	"strings"
+
+	"golang.org/x/crypto/cryptobyte"
+)
+
+// Fingerprinter is a struct largely for holding options for the FingerprintClientHello func
+type Fingerprinter struct {
+	// KeepPSK will ensure that PSK extensions are passed along into the resulting ClientHelloSpec
+	KeepPSK bool
+	// AllowBluntMimicry will ensure that unknown extensions are passed along into the resulting ClientHelloSpec
```
sergeyfrolov, line 19: I'd add "as-is"

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"golang.org/x/crypto/cryptobyte"
+)
+
+// Fingerprinter is a struct largely for holding options for the FingerprintClientHello func
+type Fingerprinter struct {
+	// KeepPSK will ensure that PSK extensions are passed along into the resulting ClientHelloSpec
```
sergeyfrolov, line 17: Still could be slightly unclear. Maybe
// KeepPSK will ensure that PSK extensions are passed along into the resulting ClientHelloSpec
// Default(false): Fingerprinter will remove the PSK from the ClientHelloSpec

u_conn_test.go
```diff
@@ -32,146 +55,332 @@ func TestUTLSMarshalNoOp(t *testing.T) {
...
+	test := &clientTest{
+		name:   "UTLS-setclienthello-" + opensslCipherName + "-" + hello.helloName(),
+		args:   []string{"-cipher", opensslCipherName},
+		config: config,
+	}
+
+	runUTLSClientTestTLS12(t, test, hello)
+}
+
+func TestUTLSHandshakeClientFingerprintedSpecFromChrome_58(t *testing.T) {
```
sergeyfrolov, line 168: This test should probably go into u_fingerprintiner_test

u_fingerprinter.go
sergeyfrolov: Meh, good enough.
Probably the right way is to create UTLS_CONTRIBUTORS file, put our past and present contributors there, and refer to that file, as I don't want to overload code headers with the names of Tom, Dick and Harry, but this is simpler and will do.
Doesn't really matter

u_fingerprinter.go
sergeyfrolov: I'd add that those issues may compromise security and mimicry

u_fingerprinter.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+			clientHelloSpec.Extensions = append(clientHelloSpec.Extensions, &KeyShareExtension{keyShares})
+
+		case extensionPSKModes:
+			// RFC 8446, Section 4.2.9
+			pskModes := []uint8{}
+			if !readUint8LengthPrefixed(&extData, &pskModes) {
+				return nil, errors.New("unable to read PSK extension data")
+			}
+			clientHelloSpec.Extensions = append(clientHelloSpec.Extensions, &PSKKeyExchangeModesExtension{pskModes})
```
sergeyfrolov, line 252: They greased the PSK modes too
https://tools.ietf.org/html/draft-ietf-tls-grease-01

u_common_test.go
```diff
@@ -0,0 +1,32 @@
...
+		isGREASE bool
+	}{
+		{0x0a0a, true},
+		{0x1a1a, true},
+		{0x2a1a, false},
+		{0x2a2a, true},
+		{0x1234, false},
+		{0x1a2a, false},
+		{0xdeed, false},
+		{0xb1b1, false},
```
sergeyfrolov, line 24: Add {0x0b0b, false} and run this again :)

u_fingerprinter_test.go
sergeyfrolov: Seems reasonable

u_fingerprinter_test.go
```diff
@@ -0,0 +1,408 @@
...
+	f = &Fingerprinter{AllowBluntMimicry: true}
+	generatedSpec, err := f.FingerprintClientHello(uconn.HandshakeState.Hello.Raw)
+	if err != nil {
+		t.Errorf("got error: %v; expected to succeed", err)
+	}
+
+	for _, ext := range generatedSpec.Extensions {
+		if genericExtension, ok := (ext).(*GenericExtension); ok {
+			if genericExtension.Id == extensionId && bytes.Equal(genericExtension.Data, extensionData) {
+				return
```
sergeyfrolov, line 256: checking just the first extension is not enough

u_common_test.go
max-b: :facepalm: :hankey: thanks for the catch. That was sloppy.

u_fingerprinter.go
max-b: I think this implementation will just re-use the GREASE values from the original hello? They have their own weird/arbitrary looking pattern, so I wasn't sure that there was consensus on how to handle them? Do you happen to know whether in general implementations use a random selector like you're using for the other GREASE values? I couldn't find anything in boringssl, but I can try and see if I can generate any locally.

Oh also - I looked at the `tls_parser.rs` file that you sent over and I _thiiiink_ it's actually not properly handling PSK modes grease? It looks like it's just applying the same `ungrease_u8` function which I don't think works for psk modes. Maybe I missed something though...

u_fingerprinter_test.go
max-b: I'm not sure I follow here - this test is to ensure that the generic extension exists in the generated spec with `AllowBluntMimicry` set. What additional checks would you want here/can you explain the kind of input this wouldn't properly handle?

u_fingerprinter.go
max-b: The current implementation returns an error if `KeepPSK` is false and it encounters an `extensionPreSharedKey`. Do you think it's preferable to just ignore it?

u_fingerprinter.go
max-b: I'm actually a bit confused as to the `TLSVersMax`/`TLSVersMin` functionality. In terms of mimicry, am I right to assume that, supposing there is no supported versions extension, we'd want to set `minTLSVers = maxTLSVers = packetVersionField`?

The only reason I hesitate w/ that approach is that for the parrots, you specify them explicitly. Does that mean that Chrome version 70 could possibly be sending v1.0 client hellos a la https://github.com/refraction-networking/utls/blob/4ee624457953e3730be8cb012f0cc0450867d1c9/u_parrots.go#L71 ?

max-b: fwiw w/ this current implementation, if the supported versions extension is present, the `ClientHelloSpec` will use those versions (because of the `SetTLSVers` function), otherwise it will default to `min = TLS 1.0, max = TLS 1.2`

u_tls_extensions.go
```diff
@@ -242,6 +242,7 @@ type ALPNExtension struct {
 func (e *ALPNExtension) writeToUConn(uc *UConn) error {
 	uc.config.NextProtos = e.AlpnProtocols
 	uc.HandshakeState.Hello.AlpnProtocols = e.AlpnProtocols
+	uc.HandshakeState.Hello.NextProtoNeg = uc.HandshakeState.Hello.NextProtoNeg || len(e.AlpnProtocols) > 0
```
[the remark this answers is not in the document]
max-b: Ah yes I should have mentioned this to you - I was meaning to ask you about it.

Currently, without that line, the `TestUTLSFingerprintClientHello` test fails when checking mimicry against the `Randomized-ALPN` spec:
```
NextProtoNeg fields not equal, expected: true, got: false
```

I _believe_ that's because when creating the randomized spec, it actually touches the underlying `UConn`'s `Config`:
https://github.com/refraction-networking/utls/blob/4ee624457953e3730be8cb012f0cc0450867d1c9/u_parrots.go#L810-L817

This in turn means that when `makeClientHello` is called, the hello's `NextProtoNeg` is set true:
https://github.com/refraction-networking/utls/blob/master/handshake_client.go#L77-L79

However, `FingerprintClientHello` doesn't actually have access to a `Config`, so it can't make the kind of parallel change.

I'm honestly struggling a tiny bit with fully understanding the interaction between `ALPN` and `NPN` re: the golang crypto implementation, but I _think_ that this makes sense. I'm *very* open to an explanation of how I'm wrong.

u_fingerprinter.go
max-b: I think it was a kind of least necessary information thing ie for tests (or perhaps other purposes??) where the caller only has the `ClientHello` message, it would require fabricating a the rest of the record?

I think at first I also chose it because of its parallels with `crypto/tls.clientHelloMsg.marshal()`.

I've gone ahead and changed it to accept the full record.

max-b:
Thanks so much for your patience with this process! I really appreciate the feedback and I suspect you are busy right now :grin: 

Let me know what you think about these changes.

Good luck tomorrow!

u_fingerprinter.go
```diff
@@ -0,0 +1,330 @@
...
+					!readUint16LengthPrefixed(&clientShares, &ks.Data) ||
+					len(ks.Data) == 0 {
+					return nil, errors.New("unable to read key share extension data")
+				}
+				ks.Group = CurveID(unGREASEUint16(group))
+				// if not GREASE, key share data will be discarded
+				// as it should be generated per connection
+				if ks.Group != GREASE_PLACEHOLDER {
+					ks.Data = nil
+				}
```
max-b, line 269: Just a heads up that a teammate pointed out that the actual key_share "data" should be generated per connection. I could make another option for this, but it seems unlikely to actually be what the end user wants?

sergeyfrolov: I think users would appreciate this option, and don't you want this in your software? Also, have you had a chance to try integrating this feature? That may reveal existing issues, and allow to make changes to the API before it's committed.

I suspect that the option we'd want would be along the lines of DisableSecretsRegeneration, such that by default it would re-randomize ClientRandom and regenerate the keys in key_share, but one can set it to true to disable that.

max-b: :+1: thanks for the point here. I don't think we're currently planning on needing it, but I did add a `DisableSecretsRegeneration` flag for future flexibility/options.

I've also added a `AlwaysAddPadding` option at the request of one of my colleagues.

max-b: >  Also, have you had a chance to try integrating this feature? That may reveal existing issues, and allow to make changes to the API before it's committed.

We're very much in the process of this :grin: I _thiiink_ we may have it in production by next week, at least for some percentage of users.

So far the API suggestions you've made have been very helpful!

u_fingerprinter.go
sergeyfrolov: Seems like we need `GREASE_PSK_PLACEHOLDER = 0x0B`, to which we would set any PSK grease value, and on re-creating the ClientHello would re-randomize it.

I think I can accept the PR with this issue as is, because:
1) Thankfully, Google Chrome does not grease PSK for some reason right now.
2) The outcome is not that bad: you will just keep using the same PSK grease value, instead of re-randomizing it, which is suspicious but doesn't seem too bad.
However, we should at least leave a comment and maybe open an issue to document this.

u_tls_extensions.go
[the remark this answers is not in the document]
sergeyfrolov: Default Golang behavior is that if config.NextProtos is not nil, it will enable both NPN and ALPN. I think it's important that we set it correctly because we may run into the issues in this chunk of code:
https://github.com/refraction-networking/utls/blob/43c36d3c1f57546d5cbb05c066df7b5a78686c51/handshake_client.go#L681-L699

The correct fix would be to look if our fingerprint actually has NPN by creating a bool `haveNPN` and then filling it when we are cycling through the extensions here:
https://github.com/refraction-networking/utls/blob/33a29038e742910d6ec82636748dfad5dd7f30ae/u_parrots.go#L621
and then `config.NextProtoNeg =  haveNPN`.
Please leave a comment why we're doing this: default makeClientHello always sets NextProtoNeg and sends the NPN extension if NextProtos, also used by ALPN, is set, but we may not actually have NPN in our ClientHelloSpec.

sergeyfrolov (changes requested):

max-b:
I realized that it's going to bug me if this just dies here and I think it's pretty close to the finish line, so I'm going to respond to your comments in this PR and then open a new one from my personal account with the requested changes.

u_tls_extensions.go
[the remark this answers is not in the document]
max-b: Great ok took me a little while to wrap my head around this, but I believe I understand what you're saying and I'll make the fix.

u_fingerprinter.go
max-b: Ah ok thanks for that bit of primer. I'll make this change.

u_fingerprinter.go
max-b: Ok I'll add the comment and I can open an issue to track this. If it becomes necessary I believe I could probably contribute the re-greasing.

u_fingerprinter_test.go
max-b: To be (hopefully) more clear: this test manually creates a single generic extension in the client hello which is to be fingerprinted. That generic extension has data = `extensionData`. This for loop is solely interested in finding that the generated spec has had that generic extension with `extensionData` included

u_fingerprinter.go
sergeyfrolov: Oh, my bad. When reading your first comment, I misunderstood you and thought that you weren't actually planning on regenerating the key_share data. Of course, regenerating it is the right behavior, I can't imagine why anyone wouldn't want to do it.
I'd even argue we should remove the DisableSecretsRegeneration. If someone really wants to do it, they actually can set whatever keys_share they want, it just would take some extra steps. I was only suggesting we add that option because I thought you wanted to not regenerate the secret for some reason, and wanted to give you the privilege of convenient API for contributing the feature.

u_fingerprinter.go
sergeyfrolov: I think so, yes.

u_fingerprinter.go
max-b: Gotcha - I've removed it now.

PR #61, opened 2020-08-19
Expose DecryptTicketWith function

max-b:
Saw that you might be open to this in https://github.com/refraction-networking/utls/pull/38/commits/47053bd8e4d51eda609484ccaf263dd8c56366fe#r348910445 so I figured I'd open it as its own PR

### DIFF
--- a/ticket.go
+++ b/ticket.go
@@ -173,11 +173,12 @@ func (c *Conn) encryptTicket(state []byte) ([]byte, error) {
 
 // [UTLS] changed to use exported DecryptTicketWith func below
 func (c *Conn) decryptTicket(encrypted []byte) (plaintext []byte, usedOldKey bool) {
-	return DecryptTicketWith(encrypted, c.config)
+	tks := ticketKeys(c.config.ticketKeys()).ToPublic()
+	return DecryptTicketWith(encrypted, tks)
 }
 
-// [UTLS] changed to be made public and take a Config struct instead of use a Conn receiver
-func DecryptTicketWith(encrypted []byte, config *Config) (plaintext []byte, usedOldKey bool) {
+// [UTLS] changed to be made public and take a TicketKeys instead of use a Conn receiver
+func DecryptTicketWith(encrypted []byte, tks TicketKeys) (plaintext []byte, usedOldKey bool) {
 	if len(encrypted) < ticketKeyNameLen+aes.BlockSize+sha256.Size {
 		return nil, false
 	}
@@ -187,7 +188,9 @@ func DecryptTicketWith(encrypted []byte, config *Config) (plaintext []byte, used
 	macBytes := encrypted[len(encrypted)-sha256.Size:]
 	ciphertext := encrypted[ticketKeyNameLen+aes.BlockSize : len(encrypted)-sha256.Size]
 
-	keys := config.ticketKeys()
+	// keys := c.config.ticketKeys() // [UTLS] keys are received as a function argument
+
+	keys := tks.toPrivate()
 	keyIndex := -1
 	for i, candidateKey := range keys {
 		if bytes.Equal(keyName, candidateKey.keyName[:]) {
--- a/ticket.go
+++ b/ticket.go
@@ -171,7 +171,17 @@ func (c *Conn) encryptTicket(state []byte) ([]byte, error) {
 	return encrypted, nil
 }
 
+// [UTLS] changed to use exported DecryptTicketWith func below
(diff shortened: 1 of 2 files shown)


ticket.go
```diff
@@ -181,7 +191,7 @@ func (c *Conn) decryptTicket(encrypted []byte) (plaintext []byte, usedOldKey boo
 	macBytes := encrypted[len(encrypted)-sha256.Size:]
 	ciphertext := encrypted[ticketKeyNameLen+aes.BlockSize : len(encrypted)-sha256.Size]
 
-	keys := c.config.ticketKeys()
```
sergeyfrolov, line 184: Comment this line out instead of removing. Something like
// keys := c.config.ticketKeys() // [uTLS] keys are received as function argument

ticket.go
```diff
@@ -171,7 +171,17 @@ func (c *Conn) encryptTicket(state []byte) ([]byte, error) {
...
 }
 
+// [UTLS] changed to use exported DecryptTicketWith func below
 func (c *Conn) decryptTicket(encrypted []byte) (plaintext []byte, usedOldKey bool) {
+	return DecryptTicketWith(encrypted, c.config)
+}
+
+// [UTLS] changed to be made public and take a Config struct instead of use a Conn receiver
+func DecryptTicketWith(encrypted []byte, config *Config) (plaintext []byte, usedOldKey bool) {
+	if len(encrypted) < ticketKeyNameLen+aes.BlockSize+sha256.Size {
```
sergeyfrolov, line 181: same check twice

max-b: :facepalm:

max-b:
>  However, why not add TicketKey to u_public.go, accept (Keys TicketKeys) directly, and then on the first line of this function cast it to private keys?

Sounds good to me - I implemented that. Let me know if it looks ok.

> Also it's not obvious what usedOldKey is.

Agreed :grin:  I attempted to improve the documentation, though perhaps you have something you prefer?


Thanks!

u_public.go
```diff
@@ -612,3 +612,52 @@ func (css *ClientSessionState) SetServerCertificates(ServerCertificates []*x509.
...
+
+func (tks ticketKeys) ToPublic() []TicketKey {
+	var TKS []TicketKey
+	for _, ks := range tks {
+		TKS = append(TKS, ks.ToPublic())
+	}
+	return TKS
+}
+
+func (TKS TicketKeys) toPrivate() []ticketKey {
```
sergeyfrolov, line 657: ToPrivate()

u_public.go
```diff
@@ -612,3 +612,52 @@ func (css *ClientSessionState) SetServerCertificates(ServerCertificates []*x509.
...
+
+func (tk ticketKey) ToPublic() TicketKey {
+	return TicketKey{
+		KeyName: tk.keyName,
+		AesKey: tk.aesKey,
+		HmacKey: tk.hmacKey,
+	}
+}
+
+func (TK TicketKey) toPrivate() ticketKey {
```
sergeyfrolov, line 641: ToPrivate()

ticket.go
```diff
@@ -171,7 +171,20 @@ func (c *Conn) encryptTicket(state []byte) ([]byte, error) {
...
+// [UTLS] changed to use exported DecryptTicketWith func below
 func (c *Conn) decryptTicket(encrypted []byte) (plaintext []byte, usedOldKey bool) {
+	tks := ticketKeys(c.config.ticketKeys()).ToPublic()
+	return DecryptTicketWith(encrypted, tks)
+}
+
+// DecryptTicketWith decrypts an encrypted session ticket
+// using a TicketKeys (ie []TicketKey) struct
+//
+// usedOldKey will be true if the key used for decryption is
```
sergeyfrolov, line 183: "DecryptTicketWith returns the decrypted session ticket, and a boolean usedOldKey,
which will be true, if the key used for decryption was not the first in the []TicketKey slice."

I don't see where usedOldKey is set though.

ticket.go
sergeyfrolov: [uTLS] is preferable

ticket.go
max-b: Yeah I thought that it was confusing the way they were using a named return there too:
https://github.com/refraction-networking/utls/blob/33a29038e742910d6ec82636748dfad5dd7f30ae/ticket.go#L213

I think maybe it was as a form of documentation?

sergeyfrolov:
Ah, somehow didn't notice that, thought the function ends with just a
return.

Issue #16, opened 2019-01-11
Using utls with `http.Transport`


AxbB36:
I'm using commit a89e7e6da482a5a0db02578fc606ace9ccfbea62. The examples I have found of using utls with HTTPS all make a single request on a single connection, then throw the connection away. For example, [`httpGetOverConn`](https://github.com/refraction-networking/utls/blob/a89e7e6da482a5a0db02578fc606ace9ccfbea62/examples/examples.go#L318) in examples.go.

I'm trying to use utls with [`http.Transport`](https://golang.org/pkg/net/http/#Transport), to take advantage of persistent connections and reasonable default timeouts. To do this, I'm hooking into the `DialTLS` callback. There is a problem when using a utls fingerprint that includes `h2` in ALPN and a server that supports HTTP/2. The server switches to HTTP/2 mode, but the client stays in HTTP/1.1 mode, because net/http [disables automatic HTTP/2 support whenever `DialTLS` is set](https://github.com/golang/go/blob/a2bb68de4d1cb1ca35279523365e9caf36fb9896/src/net/http/transport.go#L299-L306). The end result is an HTTP/1.1 client speaking to an HTTP/2 server; i.e, a similar problem as what was reported in https://github.com/golang/go/issues/14275#issue-132513041. The error message differs depending on the fingerprint:
<dl>
<dt><code>HelloFirefox_63</code></dt>
<dd><code>net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x12\x04\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00d\x00\x04\x00\x10\x00\x00\x00\x06\x00\x00@\x00\x00\x00\x04\b\x00\x00\x00\x00\x00\x00\x0f\x00\x01\x00\x00\x1e\a\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01http2_handshake_failed"</code></dt>
<dt><code>HelloChrome_70</code></dt>
<dd><code>local error: tls: unexpected message</code></dd>
<dt><code>HelloIOS_11_1</code></dt>
[...]

PR #74, opened 2021-03-23
feat: sketch out utls http.RoundTripper

bassosimone:
We have experimental OONI code that is similar to an http.RoundTripper for `utls` as described in https://github.com/refraction-networking/utls/issues/16. So, I choose to spend some time to make one. If you like the concept, then I can also write and add unit tests to this PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/refraction-networking/utls
+
+go 1.16
+
+require (
+	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
+	golang.org/x/net v0.0.0-20210323141857-08027d57d8cf
+)
--- a/u_http.go
+++ b/u_http.go
@@ -1,0 +1,562 @@
+package tls
+
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"golang.org/x/net/http2"
+	"golang.org/x/net/idna"
+	"golang.org/x/net/proxy"
+)
+
+// uhttpLogger is the type of the optional logger. The log.Default
+// logger in the standard library matches this interface.
+type uhttpLogger interface {
+	Printf(fmt string, v ...interface{})
+	Print(v ...interface{})
+}
+
+// uhttpSilentLogger is the silent logger.
+type uhttpSilentLogger struct{}
+
+// Printf implements uhttpLogger.Printf.
+func (ul *uhttpSilentLogger) Printf(fmt string, v ...interface{}) {}
+
+// Print implements uhttpLogger.Print.
+func (ul *uhttpSilentLogger) Print(v ...interface{}) {}
+
+// uhttpLog is the default logger.
+var uhttpLog uhttpLogger = &uhttpSilentLogger{}
+
+// init checks whether the user wants verbose logs.
+func init() {
+	if os.Getenv("UHTTP_VERBOSE") == "1" {
+		uhttpLog = log.Default()
+	}
+}
+
+// UHTTPTransport uses UTLS instead of TLS. This struct
+// mimicks an http.Transport and matches the http.RoundTripper
+// standard library interface.
+//
+// As documented in https://github.com/refraction-networking/utls/issues/16,
+// the standard library http.RoundTripper cannot use connections
+// from UTLS because httpTransport enables http2 only when it's possible
+// to cast the net.Conn to a *tls.Conn.
+//
+// This transport attempts to solve this issue by inspecting
+// the ALPN negotiated protocol and routing:
+//
+// - "h2" to a default constructed http2.Transport;
+//
+// - "http/1.1" to a default constructed http.Transport.
+//
+// Moreover, cleartext HTTP requests go to a default
+// constructed http.Transport.
+//
+// The zero initialized UHTTPTransport is valid and can be
+// used immediately. We will allocate internal variables when
+// we need them. As http.Transport, UHTTPTransport may have
+// idle connections, for which CloseIdleConnections can be used.
+//
+// You SHOULD NOT modify the public fields of this data
+// structure while it's being used, because that MAY
+// quite possibly lead to data races. Otherwise, it is
+// safe to call the methods of this struct from several
+// concurrent goroutines.
+type UHTTPTransport struct {
+	// TODO(bassosimone): what useful fields of an ordinary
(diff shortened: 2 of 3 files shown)


u_http.go
```diff
@@ -0,0 +1,565 @@
...
+	onlyDial uhttpCloseableTransport
+
+	// https is a transport used only for the "http/1.1"
+	// ALPN. This transport does not perform any dial and
+	// only manages its cached persistent connections. We'll
+	// initialize it during the first RoundTrip call.
+	https uhttpCloseableTransport
+
+	// h2 is like https but for the "h2" ALPN.
+	h2 uhttpCloseableTransport
```
sergeyfrolov, line 126: nitpick: I'd prefer a more telling naming scheme for those transports, something like
`h1Cleartext`
`h1Secure`
`h2Secure`

u_http.go
```diff
@@ -0,0 +1,565 @@
...
+
+// Print implements uhttpLogger.Print.
+func (ul *uhttpSilentLogger) Print(v ...interface{}) {}
+
+// uhttpLog is the default logger.
+var uhttpLog uhttpLogger = &uhttpSilentLogger{}
+
+// init checks whether the user wants verbose logs.
+func init() {
+	if os.Getenv("UHTTP_VERBOSE") == "1" {
```
sergeyfrolov, line 43: Please add a function that will let users enable verbose logging without having to set environment variables (and mention it in the readme).

u_http.go
```diff
@@ -0,0 +1,565 @@
...
+
+	// cleartext is a transport for HTTP only. We will initialize
+	// this field on the first RoundTrip invocation.
+	cleartext uhttpCloseableTransport
+
+	// onlyDial is the transport used for dialing new
+	// connections. It will populate connCache and
+	// hostCache. We will initialize this field during
+	// the first invocation of RoundTrip.
+	onlyDial uhttpCloseableTransport
```
sergeyfrolov, line 117: After a brief look I didn't understand the purpose of `onlyDial`.

go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/refraction-networking/utls
+
+go 1.16
```
sergeyfrolov, line 3: Please lower the minimum version as low as possible.

u_http.go
```diff
@@ -0,0 +1,565 @@
...
+		return nil, errors.New("uhttp: unsupported URL scheme")
+	}
+	// Step 2: check whether we have a proxy URL.
+	proxyURL, err := txp.proxy(req)
+	if err != nil {
+		return nil, err
+	}
+	// Step 3: dispatch HTTPS requests to the proper transport
+	child := txp.hostCacheGetOrDefault(req.URL)
+	const maxRetries = 4
```
sergeyfrolov, line 203: We should allow users to change maxRetries, perhaps as a field in UHTTPTransport

u_http.go
```diff
@@ -0,0 +1,565 @@
...
+	}
+	uhttpLog.Printf("uhttp: h2: connCache miss for %s", address)
+	return nil, errUHTTPNoCachedConn
+}
+
+// dialUTLSContext dials a TLS connection using UTLS and the
+// settings configured inside UHTTPTransport. This function
+// updates hostCache and saves the connection into connCache,
+// on success. Note that success is indicated by returning
+// one of errUHTTPUse{H2,HTTPS}.
```
sergeyfrolov, line 386: Let's not indicate success by returning errors. Seems like this will need some refactoring.

u_http.go
```diff
@@ -0,0 +1,565 @@
...
+		txp.cleartext.CloseIdleConnections()
+	}
+	if txp.https != nil {
+		txp.https.CloseIdleConnections()
+	}
+	if txp.h2 != nil {
+		txp.h2.CloseIdleConnections()
+	}
+	defer txp.mu.Unlock()
+	txp.mu.Lock()
```
sergeyfrolov, line 558: Don't we need to lock earlier?

sergeyfrolov (commented):
Hi Simone. I took a brief look, and I do like the concept -- many thanks for the contribution! I didn't quite figure out how caching works yet, but I left some comments there, will take another look later.
Please don't forget to add a section in README to tell people how to use this.

bassosimone:
Thanks for the quick first round of review! :-)

u_http.go
bassosimone: Sure! (The rationale of this design with the environment variable was to do what the standard library does in terms of logging: they check the `GODEBUG` environment variable to determine whether to emit debug events.)

u_http.go
bassosimone: The `http/1.1` and the `h2` transports are configured to reject any configuration attempt with an error saying that there are no cached connections (`errUHTTPNoCachedConn`). The `onlyDial` transport is only used for dialing TCP/UTLS connections. In turn, `onlyDial` is configured to create connections, store them into a cache as a side effect, and then return an error indicating whether to use `http/1.1` or `h2` based on the the ALPN value. Maybe I can explain in much more detail this fact in the documentation of `onlyDial`?

u_http.go
bassosimone: Sure!

go.mod
bassosimone: Sure! I suppose the minimum required version is Go 1.14 where `DialTLSContext` was added to `http.Transport`. (I am also using `log.Default`, which is Go >= 1.16, but that functionality is definitely not fundamental to this patch.)

u_http.go
bassosimone: Good point. Side effects and errors as sentinels are bad. Thank you for your comment, this is really what code review is about! I suppose there is here a good opportunity for simplifying the codebase by getting rid of `onlyDial` and running the `dialUTLSContext` right inside `RoundTrip`. By running the diff on memory, this seems possible to do and it also seems to yield a simpler algorithm inside of `RoundTrip`, which is also good.

u_http.go
bassosimone: If goroutine A calls `RoundTrip` and goroutine B calls `CloseIdleConnections` concurrently, we have a data race because the write on, say, `txp.h2` is not guaranteed to happen before the read that we do here.

(I would still like the `CloseIdleConnections` to go unlocked, because it does not need to synchronize with any other synchronized access performed by this structure, but I should refactor the code to avoid this race.)

bassosimone:
Okay, @sergeyfrolov, I've gone through the comments, and it seems to me the bottom line here is that there is an opportunity for refactoring to avoid side effects, the need to reorganize locking in `CloseIdleConnections` and a bunch of nitpicks. Will do that!

PR #76, opened 2021-04-07
Support for TLS1.2 SessionID Based Resumption

DunyaKokoschka:
Add support for TLS1.2 session ID based resumption. Some clients notably iOS NSURLSession by default only support session id based resumption for TLS1.2 so it would be useful to be able to support session id based resumption.

This adds SessionIdEnabled which defaults to false to preserve backwards compatibility and ServerSessionCache for storing sessionId based sessions to the tls Config object. We do not supply an implementation of ServerSessionCache like the LRU client cache. I am not particularly interested in getting session id based resumption working for the server and only have provided an implementation so the client implementation can easily be tested.

The changes to the normal Golang TLS server/client seem straight forward but I'm not so sure about the uTLS changes in particular ApplyPresets which now unconditionally calls SetSessionState whereas previously it only called SetSessionState if the SessionTicketExtension was present. Also, the SetSessionCache method now does not make sense since it unconditionally sets `TicketSupported = true` even though the `SessionCache` can now be used without session tickets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ Some customizations(such as setting session ticket/clientHello) have easy-to-use
     sessionState := utls.MakeClientSessionState(sessionTicket, uint16(tls.VersionTLS12),
         tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
         masterSecret,
-        nil, nil)
+        nil, nil, nil)
     tlsConn.SetSessionState(sessionState)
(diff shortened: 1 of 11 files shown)


Gaukas (approved):

gaukas:
Closing old PRs that timed out. Also I believe this has been addressed when we revisit PSK-based resumptions. But still, thank you!

PR #102, opened 2022-03-07
Add wasm support

arriven:
Fixes #71

PR #106, opened 2022-05-16
add loong64 support 

cheungxi:
loong64 (loongarch64) a new isa

### DIFF
--- a/cpu/cpu_loong64.go
+++ b/cpu/cpu_loong64.go
@@ -1,0 +1,13 @@
+// Copyright 2022 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build loong64
+// +build loong64
+
+package cpu
+
+const cacheLineSize = 64
+
+func initOptions() {
+}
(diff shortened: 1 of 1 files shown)


yetist:
@gaukas 

Please review this patch, otherwise I have to use it like this:
```
    go mod edit -replace=github.com/refraction-networking/utls=github.com/cheungxi/utls@master
    go mod tidy                                                                  
```
Thanks.

gaukas:
Hi @yetist, 

I don't really have any major blocker to get this merged. But since `loong64` is not even included in `go/internal`, I'd like to know how are you using other packages on this ISA and what is the specific usage of using uTLS on it. 

I guess that's a very personal question, feel free not to respond ;) 

And since this PR doesn't really introduce any new function or urgent fix, it might not make a new PR but rather get combined into the next release. Thanks. 

And thanks for @cheungxi for contributing!

gaukas:
And also I don't really have a good way to test on different ISAs, that's a reason I was not merging either this PR or #102. Before I go ahead and merge it, have you checked that this would work instead of breaking anything? Just to make sure.

VeNoMouS:
Just to clarify... this is to support the cpu arch related to https://github.com/golang/go/issues/46229 ?

gaukas:
Yes, I believe they are referring to the same architecture by loong64.

Issue #113, opened 2022-08-07
Can you help support TLS Extension application_settings(17513)


deancn:
Extension: application_settings (len=5)
      Type: application_settings (17513)
    
    
https://www.ietf.org/archive/id/draft-vvv-tls-alps-01.html

r3nby:
I second this.  The UTLS client is missing the 17513 extension which causes it to get 403'd on some websites (cloudflare).

```
// chrome 96 wireshark
// JA3 Hash: cd08e31494f9531f560d64c695473da9
// JA3 Full: 771,4865-4866-4867-49195-49199-49196-49200-52393-52392-49171-49172-156-157-47-53,0-23-65281-10-11-35-16-5-13-18-51-45-43-27-17513-21,29-23-24,0

// chrome_83 client hello golang
// JA3 Hash: b32309a26951912be7dba376398abc3b
// JA3 Full: 771,4865-4866-4867-49195-49199-49196-49200-52393-52392-49171-49172-156-157-47-53,0-23-65281-10-11-35-16-5-13-18-51-45-43-27-21,29-23-24,0

```

Wireshark Client Hello for chrome 96.

```
Frame 49: 571 bytes on wire (4568 bits), 571 bytes captured (4568 bits) on interface enp4s0, id 0
Ethernet II, Src: ASRockIn_25:78:e4 (a8:a1:59:25:78:e4), Dst: Netgear_a2:e7:54 (6c:cd:d6:a2:e7:54)
Internet Protocol Version 4, Src: 192.168.1.7, Dst: 104.18.193.78
Transmission Control Protocol, Src Port: 59276, Dst Port: 443, Seq: 1, Ack: 1, Len: 517
Transport Layer Security
    TLSv1.2 Record Layer: Handshake Protocol: Client Hello
        Content Type: Handshake (22)
        Version: TLS 1.0 (0x0301)
        Length: 512
        Handshake Protocol: Client Hello
            Handshake Type: Client Hello (1)
            Length: 508
            Version: TLS 1.2 (0x0303)
            Random: 632375267fb762a679c020653e0b5fee5af889998e602608d0e4343648bd3f1b
            Session ID Length: 32
            Session ID: 737bb0d18030f5f0b47a0e4525137583a3bb4036e7a29dad7dc596c891e9e974
            Cipher Suites Length: 32
            Cipher Suites (16 suites)
            Compression Methods Length: 1
            Compression Methods (1 method)
            Extensions Length: 403
[...]

Issue #108, opened 2022-06-04
ApplicationSettingsExtension


kawacode:
Hey,
ApplicationSettingsExtension is needed!
https://github.com/Noooste/utls supports it, so you could maybe take a look.
Thx

Issue #107, opened 2022-06-04
feature: extensionPreSharedKey support


kawacode:
Hey,
is there a way in Utls to add PSK/pre_shared_key/extensionPreSharedKey,
all i can find in utls is that somewhere you guys say: TODO: wait for someone who cares about PSK to implement
Thanks.

yuzd:
+1

gaukas:
Seemingly our upstream `crypto/tls` is claimed to have enough support implemented for `pre_shared_key`. 

Since all changes up to Go 1.21.0 has been integrated into uTLS, it should be possible to implement support for `PreSharedKeyExtension` based off `FakePreSharedKeyExtension`.

jjsaunier:
- device/user authentication through PSK (out of band)
- In proxy app, leverage 0-RTT resumption
- Still in proxy app, leverage server-side state https://datatracker.ietf.org/doc/html/rfc5077 
- In TLS mitm, replicate 1:1 the incoming fingerprint (this is already ok) and n+1 handshake behavior (PSK behavior is not replicated and forwarded, it redo a full handshake)

gaukas:
If any of you are familiar with the PSK process implemented in `crypto/tls`, please report here what internal/external state/variable must be set to enable PSK-based resumption on Client. 

So far I tried setting `config.ClientSessionCache` but it doesn't work, the client doesn't advertise support for session_ticket and therefore receives no `New Session Ticket` message.

gaukas:
So yesterday I spent some time on `crypto/tls` to play with it a bit. Seemingly simply setting `ClientSessionCache` would work but you will need to do a full handshake **plus** reading at least one post-handshake message from the TLSConn to trigger the handling of a NewSessionTicket. 

Made a few changes to uTLS, so at least it can calculate the binder correctly. However something could be missing, as the local key is not updated properly and we are getting `local error: tls: bad record MAC`

gaukas:
#231 adds support for PSK.

3andne:
Hi @gaukas,
Nice work on the PSK support. However I got some questions:
> FakePreSharedKeyExtension uses CipherSuiteID, SessionSecret and Identities to calculate the corresponding binders and send them, without setting the internal states. Therefore if the server accepts the PSK and tries to resume, the connection fails.

Why doesn't it set the internal states? Is this a non-goal or just not being a high priority? 
[...]

PR #115, opened 2022-08-24
Document adding custom fake TLS extensions

fedosgad:
I see a lot of people asking to add support for a specific extensions (for example, #113, #108, #107). However, the need is often to match required TLS ClientHello fingerprint and not to actually implement functionality required by such extensions. uTLS already has `Fake*` extensions which only serve as a placeholders, and `GenericExtension`, which (as stated in comments) "allows to include in ClientHello arbitrary unsupported extensions".

Thus I'd like to contribute documentation on creating such extensions.


gaukas:
I would love to get this merged! Thanks for updating the documentation!

PR #117, opened 2022-09-13
Maintenance: sync with upstream crypto/tls

gaukas:
Sync with latest implementation from upstream `crypto/tls` and make some minor changes to fix the incompatibility. 

## Introduced new issues
- All test cases in `TestUTLSHandshakeClientParrotGolang` has failed. 
    - It is possibly OK, since `crypto/tls` has been modified heavily and we won't expect to see the server response to be the same as 4 years ago. 
    - The failing tests are now disabled. But contributors are welcome to fix these tests. 

## Persistent issues
- The `forgeConn()` in the example in `examples` folder is broken, since the cipher suite checking in `tls.MakeConnWithCompleteHandshake()` doesn't work with **TLS 1.3**.


gaukas:
Possibly want to get this merged fast, before any merge conflict arises.

gaukas:
Will provide an alternative version with 2 commits: one for sync up and one for modifications needed for compatibility.

gaukas:
See #120 instead.

PR #120, opened 2022-09-26
Sync upstream crypto/tls

gaukas:
Including all commits to `crypto/tls` since fork happened. 

Provided with fixing commits for compatibility.


ewust (approved):
- Don't comment out broken tests (fix them, or remove them if they're not valid / don't make sense)
- Okay to tag a version/commit before we update, so we don't have to maintain backward compatibility with older go versions
- Fix the broken TestUTLSHandshakeClientParrotGolang tests?

PR #121, opened 2022-09-28
Bugfix: FingerprintClientHello should work when the dump contains the…

svmk:
… client's greeting and subsequent frames. Lack of subsequent frames should not lead to inoperability of FingerprintClientHello.


gaukas:
Thanks a ton! Let's get it merged.

PR #122, opened 2022-10-08
Add new ClientHellos

hwh33:
Also add faked support for token binding, ALPS, and delegated credentials.

This add new `ClientHelloSpec` based on:
- Firefox 105
- Chrome 106
- Edge 106
- Safari 16.0
- 360 Secure Browser 11.0
- QQ Browser 11.1

Unfortunately, the specs based on Edge 106 and 360 11.0 seem to incompatible with this library.  I haven't had time yet to investigate why.  I added the specs anyway in case they are useful to some people.  Perhaps they will function in limited use cases.  I did not update `HelloEdge_Auto` and `Hello360_Auto` to avoid breaking code dependent on those references.

### DIFF
--- a/common.go
+++ b/common.go
@@ -96,7 +96,6 @@ const (
 	extensionSignatureAlgorithmsCert uint16 = 50
 	extensionKeyShare                uint16 = 51
 	extensionNextProtoNeg            uint16 = 13172 // not IANA assigned
-	extensionALPS                    uint16 = 17513
 	extensionRenegotiationInfo       uint16 = 0xff01
 )
 
--- a/u_common.go
+++ b/u_common.go
@@ -23,7 +23,11 @@ const (
 	utlsExtensionCompressCertificate uint16 = 27
 
 	// extensions with 'fake' prefix break connection, if server echoes them back
-	fakeExtensionChannelID uint16 = 30032 // not IANA assigned
+	fakeExtensionTokenBinding         uint16 = 24
+	fakeExtensionChannelIDOld         uint16 = 30031 // not IANA assigned
+	fakeExtensionChannelID            uint16 = 30032 // not IANA assigned
+	fakeExtensionALPS                 uint16 = 17513 // not IANA assigned
+	fakeExtensionDelegatedCredentials uint16 = 34
 
 	fakeRecordSizeLimit uint16 = 0x001c
 
@@ -44,15 +48,24 @@ const (
 
 	FAKE_TLS_DHE_RSA_WITH_AES_128_CBC_SHA    = uint16(0x0033)
 	FAKE_TLS_DHE_RSA_WITH_AES_256_CBC_SHA    = uint16(0x0039)
-	FAKE_TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 = uint16(0x009f)
 	FAKE_TLS_RSA_WITH_RC4_128_MD5            = uint16(0x0004)
+	FAKE_TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 = uint16(0x009f)
+	FAKE_TLS_DHE_DSS_WITH_AES_128_CBC_SHA    = uint16(0x0032)
+	FAKE_TLS_DHE_RSA_WITH_AES_256_CBC_SHA256 = uint16(0x006b)
+	FAKE_TLS_DHE_RSA_WITH_AES_128_CBC_SHA256 = uint16(0x0067)
 	FAKE_TLS_EMPTY_RENEGOTIATION_INFO_SCSV   = uint16(0x00ff)
+
+	// https://docs.microsoft.com/en-us/dotnet/api/system.net.security.tlsciphersuite?view=netcore-3.1
+	FAKE_TLS_ECDHE_ECDSA_WITH_3DES_EDE_CBC_SHA = uint16(0xc008)
 )
 
 // newest signatures
 var (
 	FakePKCS1WithSHA224 SignatureScheme = 0x0301
 	FakeECDSAWithSHA224 SignatureScheme = 0x0303
+
+	FakeSHA1WithDSA   SignatureScheme = 0x0202
+	FakeSHA256WithDSA SignatureScheme = 0x0402
 
 	// fakeEd25519 = SignatureAndHash{0x08, 0x07}
 	// fakeEd448 = SignatureAndHash{0x08, 0x08}
@@ -109,6 +122,10 @@ const (
 	helloChrome           = "Chrome"
 	helloIOS              = "iOS"
 	helloAndroid          = "Android"
+	helloEdge             = "Edge"
+	helloSafari           = "Safari"
+	hello360              = "360Browser"
+	helloQQ               = "QQBrowser"
 
 	// versions
 	helloAutoVers = "0"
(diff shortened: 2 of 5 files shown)


hwh33:
There's been some discussion about recent GWF blocking in https://github.com/net4people/bbs/issues/129.  In that thread, there was discussion about TLS fingerprinting and this library was specifically called out as a means to combat this.  I can't say for sure that these latest ClientHellos will combat any fingerprinting the GFW is doing, but they should at least help clients look more like normal browser traffic.

u_fingerprinter.go
```diff
@@ -351,7 +406,7 @@ func (f *Fingerprinter) FingerprintClientHello(data []byte) (*ClientHelloSpec, e
 			} else if f.AllowBluntMimicry {
 				clientHelloSpec.Extensions = append(clientHelloSpec.Extensions, &GenericExtension{extension, extData})
 			} else {
-				return nil, fmt.Errorf("unsupported extension %#x", extension)
+				return nil, fmt.Errorf("unsupported extension %d", extension)
```
hwh33, line 409: Drive-by: TLS extension IDs are generally referenced as decimal

hwh33:
The diff is a little mangled for `u_parrots.go`.  I didn't modify existing hellos, I only added new ones.

gaukas:
Happy to get this merged. 

But does Edge 106 (also other browsers) use a different fingerprint than one of these existing Chrome fingerprint? I'm under the impression that it is based on Chromium.

hwh33:
> But does Edge 106 (also other browsers) use a different fingerprint than one of these existing Chrome fingerprint? I'm under the impression that it is based on Chromium.

Yeah, that's a good question.  I haven't checked equality of these new hellos against any of the Chrome hellos beyond 83.  I did this work on a fork; I didn't realize some newer hellos had been integrated upstream.  You are correct that Edge is based on Chromium (and 360 and QQ IIRC).

I don't have time to check equality right now, but I can try to do so in the next couple of days.

gaukas:
I will run a check later based on your fork and let you know the result, and we can go from there.

gaukas:
I think `HelloChrome_106` duplicates `HelloChrome_102`. For compatibility consideration, I would recommend not including `HelloChrome_106` for now. 

Otherwise, all good. All ClientHello fingerprints are unique.

u_common.go
```diff
@@ -164,6 +182,7 @@ var (
 	HelloChrome_96   = ClientHelloID{helloChrome, "96", nil}
 	HelloChrome_100  = ClientHelloID{helloChrome, "100", nil}
 	HelloChrome_102  = ClientHelloID{helloChrome, "102", nil}
+	HelloChrome_106  = ClientHelloID{helloChrome, "106", nil}
```
gaukas, line 185: Since `Chrome_106` shares the fingerprint with `Chrome_102`, let's keep `102` and get rid of `106`.

u_common.go
```diff
@@ -146,15 +163,16 @@ var (
...
 	HelloFirefox_55   = ClientHelloID{helloFirefox, "55", nil}
 	HelloFirefox_56   = ClientHelloID{helloFirefox, "56", nil}
 	HelloFirefox_63   = ClientHelloID{helloFirefox, "63", nil}
 	HelloFirefox_65   = ClientHelloID{helloFirefox, "65", nil}
 	HelloFirefox_99   = ClientHelloID{helloFirefox, "99", nil}
 	HelloFirefox_102  = ClientHelloID{helloFirefox, "102", nil}
+	HelloFirefox_105  = ClientHelloID{helloFirefox, "105", nil}
 
-	HelloChrome_Auto = HelloChrome_102
+	HelloChrome_Auto = HelloChrome_106
```
gaukas, line 175: If we are getting rid of `106`, `Chrome_Auto` should fall back to `Chrome_102`

gaukas (changes requested):

hwh33:
Thanks for running those checks @gaukas!  I updated according to your suggestions.

gaukas (approved):

gaukas:
Thanks @hwh33! 
And just one more thing before I get this merged... If I understand correctly, I think `ApplicationSettingsExtension` duplicates with the newly added `FakeALPSExtension`? Since we are not really implementing any functional support for `ApplicationSettingsExtension`, I think we'd rather keep the `FakeALPSExtension` and get rid of the former one. What do you think?

hwh33:
Oh, I didn't notice `ApplicationSettingsExtension`.  Hmm I agree that `FakeALPSExtension` is a better name since, as you say, there is no functional support for ALPS.  However, removing `ApplicationSettingsExtension` would be a breaking change for users of this library.

I would recommend keeping the existing type and adding a note to indicate that it is a faked extension.  I've done this here, in a separate commit: https://github.com/refraction-networking/utls/pull/122/commits/432a2f0174ebdc4adc04c3a9a9b60218f80e3d49.  What do you think?

I can squash commits before merging or keep both for posterity.  Let me know which you would prefer.

common.go
```diff
@@ -96,7 +96,6 @@ const (
 	extensionSignatureAlgorithmsCert uint16 = 50
 	extensionKeyShare                uint16 = 51
 	extensionNextProtoNeg            uint16 = 13172 // not IANA assigned
-	extensionALPS                    uint16 = 17513
```
hwh33, line 99: This library generally does not modify the standard lib files.  I removed this in favor of the `fakeExtensionALPS` value I added to `u_common.go`.

gaukas: Looks good. Thanks!

gaukas:
I'm think we may instead keep an alias. 
```
type ApplicationSettingsExtension = FakeALPSExtension
```
or the other way. 

I wonder, are there any difference between `FakeALPSExtension` and `ApplicationSettingsExtension` in your last commit, except for their naming?

gaukas:
But I guess that's another issue. How about this: I will get this merged, and we fix that issue with a new PR. 
How does this sound to you?

gaukas:
> They changed in Chrome 98 to only include tls versions 1.3 and 1.2.

Nice catch! I think we could make a change if that is the case... But anyways, in the latest Chrome 102's fingerprint we are only including TLS 1.3 and TLS 1.2 in supported versions.

> encrypted extensions in our branch

Interesting. Does `crypto/tls` implement such? If so we still have a pending upstream synchronization. If not... we always welcome PRs! 

> Google sites will reject uTLS without this implementation.

Interesting observation. I thought Google was rejecting uTLS simply because we didn't implement ALPS extension.

blakebyrnes:
> Nice catch! I think we could make a change if that is the case... But anyways, in the latest Chrome 102's fingerprint we are only including TLS 1.3 and TLS 1.2 in supported versions.



Oh, I totally missed 102 since it's out of order with the rest of the Chromes! My bad.

> Interesting. Does crypto/tls implement such? If so we still have a pending upstream synchronization. If not... we always welcome PRs!

crypto/tls implements handshake messages.. which is where some of this code has to live. My point is that uTLS hasn't created a structure like the u_conn and u_common whereby connection state, and new handshake messages can be added. That means this codebase is going to get very messy for things like Compressed Certificates and Encrypted Extensions as long as there's no pattern. I'm honestly not sure the right pattern here.. I have a little bit of a hard time following the current flow.

>Interesting observation. I thought Google was rejecting uTLS simply because we didn't implement ALPS extension.

Sort of the same thing. ALPS "implementation" is to respond to other handshake messages initiated by the server. It's a movement of "TLS extensions" into an encrypted follow-up handshake message, instead of broadcasting in the clear in the clientHello.

PR #125, opened 2022-10-12
Fix Google Parrots

### LINKED ISSUE #123: Chrome 98+ Fingerprint Mismatch
>Sorry to weigh in late on this, but the Chrome signature for 98+ is a tiny bit wrong. They changed in Chrome 98 to only include tls versions 1.3 and 1.2. They removed 1.1 and 1.0. Here's a link to the signature:
https://github.com/ulixee/utls/commit/e8f6488351f42c59042f850895cbc9df000e551c

>We have a suite that automates testing tls signatures across browsers and operating systems here, if it's of use:
*test suite:* https://github.com/unblocked-web/double-agent/tree/41489d6f260cc3861574a0f606e99d04b3bfe799
*data:* https://github.com/unblocked-web/browser-profile-data/tree/main/profiles

_Originally posted by @blakebyrnes in https://github.com/refraction-networking/utls/issues/122#issuecomment-1276156529_

gaukas:
- Move `HelloChrome_102` to the end of `HelloChrome_` fingerprints for consistency.
- Fixed #123 by adjusting mismatching SupportedVersionsExtension value in `HelloChrome_100`.


gaukas:
Even with the above patch, Chrome 100's fingerprint doesn't look right. Either way will not be well-known to tlsfingerprint.io. 

According to https://tlsfingerprint.io/id/e47eae8f8c4887b6, Chrome 100 appears to use the same fp with 102+.

PR #139, opened 2022-11-09
feat: allow custom GREASE value/body

### LINKED ISSUE #128: Impersonate ClientHello with UtlsGREASEExtension
Hello

I want to impersonate a request using two Grease Extensions, but when I try to mimic it, a different type value is generated, resulting in a different JA3 fingerprint. 

What I do in code:
`&tls.UtlsGREASEExtension{
			Value: 60138,
		}`

But in wireshark it shows different value with every request
Screenshot from 2022-10-25 11-10-55

Am I doing something wrong?

gaukas:
Added a `Customized` flag to avoid overwriting by `ApplyPreset()`.

Fixes #128.

### DIFF
--- a/u_fingerprinter.go
+++ b/u_fingerprinter.go
@@ -402,7 +402,7 @@ func (f *Fingerprinter) FingerprintClientHello(data []byte) (*ClientHelloSpec, e
 
 		default:
 			if isGREASEUint16(extension) {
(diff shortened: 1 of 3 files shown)


gaukas:
Should just use `GenericExtension` for this purpose.

PR #141, opened 2022-11-11
Refactor: Allow customization of handshake messages

gaukas:
- Split most of changes for `CompressCertExtension` made to `crypto/tls` files out and moved them to `u_` files.
- Edited some `crypto/tls` files to achieve better programmability for uTLS.
- Minor styling fix.

Opened #142 implementing ALPS extension which is based on this PR. 

Therefore, we might want to merge #142 directly instead.

Issue #140, opened 2022-11-09
v1.2.0 Refactor the code and allow custom handshake messages


gaukas:
We are planning to do the first major refactorization to improve the programmability, in order to support `ALPS` and more future extensions to come. 

More details to come... ~This may lead to a major version bump (`v1` -> `v2`)~ We are trying to minimize the incompatibility and not planning for a major version bump. 

- [x] #133 
- [x] #138  
- [x] #142

packman80:
Great changes! Thank you @gaukas

PR #142, opened 2022-11-11
feat: Implement ALPS extension

### LINKED ISSUE #124: Implement Encrypted Extensions
>Also potentially of note: we did an implementation of encrypted extensions in our branch. I don't think I can do a PR to your codebase though without a good bit of thought put into how additional handshake messages should be integrated into this project. Right now, you have to scatter them throughout the tls codebase. We pretty much just ported in from the BoringSSL test runner, but I didn't port in server side or unit tests, so that would be needed on top of this. 
https://github.com/ulixee/utls/commit/8d0f1c40c755c9b68f3d8bc926d4934f8ac4286e

>FYI: Google sites will reject uTLS without this implementation.

_Originally posted by @blakebyrnes in https://github.com/refraction-networking/utls/issues/122#issuecomment-1276156529_

gaukas:
Implemented support for `ApplicationSettingsExtension` -- now the handshake message could be recognized! 

PR based on #141. Fixes #124 and #140.

Credits: Major implementation work is done by @blakebyrnes in `ulixee/utls`.

### DIFF
--- a/u_handshake_messages.go
+++ b/u_handshake_messages.go
@@ -71,6 +71,7 @@ type utlsClientEncryptedExtensionsMsg struct {
 	raw                    []byte
 	applicationSettings    []byte
 	hasApplicationSettings bool
+	customExtension        []byte
 }
 
 func (m *utlsClientEncryptedExtensionsMsg) marshal() (x []byte) {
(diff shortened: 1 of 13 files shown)


u_handshake_messages.go
```diff
@@ -47,3 +51,83 @@ func (m *compressedCertificateMsg) unmarshal(data []byte) bool {
 	}
 	return true
 }
+
+type utlsEncryptedExtensionsMsgExtraFields struct {
```
ewust, line 55: Should we be exporting these fields? Is it something a using application might want to be able to view?

gaukas: I don't believe so. `customExtension` is just a fake place holder that isn't in any published draft/standard, while `applicationSettings` could be set in `Config` for client and read from `PeerApplicationSettings` in `ConnectionState`.

jmwample (approved):

PR #148, opened 2022-12-14
Properly check GREASE and add FakeEncryptThenMACExtension

gfw-report:
This pull request contains two commits:

First, when `AllowBluntMimicry` is set to true, support fake extension: EncryptThenMAC (22) https://datatracker.ietf.org/doc/html/rfc7366

---

Second, properly check if an extension/curve/cipherSuit is GREASE. 

The current version does not properly check GREASE in many cases. In particular, it compares an extension ID, a curve ID, or a cipherSuit against a constant value `GREASE_PLACEHOLDER` `(0x0a0a)`, without unGREASEing it first.

To fix this problem, we simply use the existing function: `isGREASEUint16()` with necessary casting.

For the records, some related error messages prior to this bug fix are:

* "applying generated spec failed: uTLS does not support 0xBABA as max version"
* "applying generated spec failed: unsupported Curve in KeyShareExtension: CurveID(56026).To mimic it, fill the Data(key) field manually"

### DIFF
--- a/u_fingerprinter.go
+++ b/u_fingerprinter.go
@@ -277,7 +277,7 @@ func (f *Fingerprinter) FingerprintClientHello(data []byte) (*ClientHelloSpec, e
 				ks.Group = CurveID(unGREASEUint16(group))
 				// if not GREASE, key share data will be discarded as it should
 				// be generated per connection
-				if ks.Group != GREASE_PLACEHOLDER {
+				if !isGREASEUint16(group) {
 					ks.Data = nil
 				}
 				keyShares = append(keyShares, ks)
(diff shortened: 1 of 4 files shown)


ewust (commented):
It would be nice to add a unit test if possible to check for this, but otherwise looks good!

gaukas:
Hi @gfw-report, 

Currently I am working on a change regarding the `Fingerprinter` (partially implementing PSK extension) and I will try to merge this into my PR.

gfw-report:
That would be great! Thanks for all your amazing work, @gaukas !

u_fingerprinter.go
```diff
@@ -277,7 +277,7 @@ func (f *Fingerprinter) FingerprintClientHello(data []byte) (*ClientHelloSpec, e
 				ks.Group = CurveID(unGREASEUint16(group))
 				// if not GREASE, key share data will be discarded as it should
 				// be generated per connection
-				if ks.Group != GREASE_PLACEHOLDER {
+				if !isGREASEUint16(group) {
```
gaukas, line 280: Is there a specific reason for this change? Since we have already saved the `unGREASEUint16(group)` to `ks.Group`, checking if `ks.Group == GREASE_PLACEHOLDER` is equivalent to checking `isGREASEUint16(group)` if no bug.

gaukas:
Changes from this PR are included in #168. Once it gets merged, I will close this PR.

PR #155, opened 2023-01-30
Add randomized clientHello generation in UTLSIdToSpec.

fedosgad:
`UTLSIdToSpec` allows user to get `ClientHelloSpec` for all fixed fingerprints, but randomized `ClientHelloSpec` cannot be generated that way - all randomization happens inside `UConn`'s private methods. As a library user, I'd like to have a way to get randomized clientHello directly, without having to create `UConn`.
This pull requests adds such functionality.

### DIFF
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -1844,6 +1844,9 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
 				},
 			},
 		}, nil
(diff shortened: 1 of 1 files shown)


gaukas:
I don't see any reason not to add this feature. Nice work and thanks!

gaukas (approved):

RPRX:
Nice work! But there is still a limitation, see https://github.com/refraction-networking/utls/pull/157

PR #157, opened 2023-02-04
Support randomized with non-nil seed in UTLSIdToSpec

RPRX:
CC @fedosgad @gaukas

### DIFF
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -1844,10 +1844,11 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
 				},
 			},
 		}, nil
-	case HelloRandomized, HelloRandomizedALPN, HelloRandomizedNoALPN:
-		// Use empty values as they can be filled later by UConn.ApplyPreset or manually.
-		return generateRandomizedSpec(id, "", nil, nil)
 	default:
+		if id.Client == helloRandomized || id.Client == helloRandomizedALPN || id.Client == helloRandomizedNoALPN {
+			// Use empty values as they can be filled later by UConn.ApplyPreset or manually.
+			return generateRandomizedSpec(id, "", nil, nil)
+		}
 		return ClientHelloSpec{}, errors.New("ClientHello ID " + id.Str() + " is unknown")
 	}
 }


RPRX:
In one of Xray-core's use cases, I want to generate a randomized Client Hello that supports TLSv1.3 and continually reuse it until the program exits. It's convenient to generate a randomized `ClientHelloSpec` and then check the `TLSVersMax` field, but using that `ClientHelloSpec` directly in many concurrent conns is not safe enough. Unfortunately, the actual `Seed` is unacquirable.

So with this change, I'm able to set the `Seed` field in advance and store it once a Client Hello that supports TLSv1.3 is generated:

```go
func init() {
	randomized := utls.HelloRandomized
	for {
		randomized.Seed, _ = utls.NewPRNGSeed()
		clientHelloSpec, _ := utls.UTLSIdToSpec(randomized)
		if clientHelloSpec.TLSVersMax == utls.VersionTLS13 {
			PresetFingerprints["randomized"] = &randomized
			break
		}
	}
}
```

gaukas:
Looks interesting (and legit). We should merge this to provide maximized flexibility. Thanks! 

Would you like to elaborate how does this change benefit circumvention tools like Xray? IMO it kinda defeats the purpose of mimicking, and makes it a fingerprint-able behavior. Unless there are some well-known popular tools with the same or similar behavior?

gaukas (approved):

RPRX:
> Would you like to elaborate how does this change benefit circumvention tools like Xray? IMO it kinda defeats the purpose of mimicking, and makes it a fingerprint-able behavior. Unless there are some well-known popular tools with the same or similar behavior?

Actually, `randomized` itself **is fingerprint-able already**, and we should avoid using it unless we have to (e.g. in some cities in Iran).
1. Sending **different unusual** Client Hellos is a very strange behavior, as normal applications will **not** behave like that.
2. There are **many preset weights** in `generateRandomizedSpec()`, but most of them **stopped changing** since https://github.com/refraction-networking/utls/commit/fd72b83e043e8ed28e5ce8cdca84a1080927c208, **2019**.

> if r.FlipWeightedCoin(0.7) {
> if r.FlipWeightedCoin(0.4) {
> p.CipherSuites = removeRandomCiphers(r, shuffledSuites, 0.4)
> if r.FlipWeightedCoin(0.63) {
> if r.FlipWeightedCoin(0.59) {
> if r.FlipWeightedCoin(0.51) || p.TLSVersMax == VersionTLS13 {
> if r.FlipWeightedCoin(0.9) {
> if r.FlipWeightedCoin(0.71) || p.TLSVersMax == VersionTLS13 {
> if r.FlipWeightedCoin(0.46) {
> if r.FlipWeightedCoin(0.62) || p.TLSVersMax == VersionTLS13 {
> if r.FlipWeightedCoin(0.74) {
> if r.FlipWeightedCoin(0.46) {
> if r.FlipWeightedCoin(0.75) {
> if r.FlipWeightedCoin(0.77) {
> if r.FlipWeightedCoin(0.25) {
> if r.FlipWeightedCoin(0.33) {

**After collecting enough amount of unusual Client Hellos**, certainly the censors will find out that they are `randomized` in uTLS.
Especially, the preset weight `0.4` for `p.TLSVersMax = VersionTLS13` is obviously unreasonable **nowadays in 2023**.

In my opinion, changing that weight to `1.0` is helpful for us to **expose less preset weights**, and certainly we will have a **more secure connection and less RTT**, and actually it's getting more and more **reasonable**, because it's very normal to support TLSv1.3 nowadays. Besides, both XTLS Vision and REALITY (a real-time Server Hello stealer + MitM, coming soon) require TLSv1.3.

By the way, these two problems will all be gone if we reuse the same `Seed` all the time. It's the most appropriate usage, probably.

gaukas:
Thanks for the detailed analysis and suggestions. 

I agree that some designs included in uTLS have been outdated and needs a lot of rework. Since it is open-sourced, you are always welcome to submit PR if you'd like to!

PR #163, opened 2023-02-06
Customable weights used in generateRandomizedSpec()

RPRX:
As described in https://github.com/refraction-networking/utls/pull/157#issuecomment-1417156797, allowing users to set their own weights would be helpful (and convenient for some purposes too).

CC @gaukas

### DIFF
--- a/u_common.go
+++ b/u_common.go
@@ -216,6 +216,26 @@ var (
 	HelloQQ_11_1 = ClientHelloID{helloQQ, "11.1", nil}
 )
 
+// Weights used in generateRandomizedSpec(), they should be set in advance.
+var (
+	Weight_Extensions_Append_ALPN                             = 0.7
+	Weight_TLSVersMax_Set_VersionTLS13                        = 0.4
+	Weight_CipherSuites_Remove_RandomCiphers                  = 0.4
+	Weight_SigAndHashAlgos_Append_ECDSAWithSHA1               = 0.63
+	Weight_SigAndHashAlgos_Append_ECDSAWithP521AndSHA512      = 0.59
+	Weight_SigAndHashAlgos_Append_PSSWithSHA256               = 0.51
+	Weight_SigAndHashAlgos_Append_PSSWithSHA384_PSSWithSHA512 = 0.9
+	Weight_CurveIDs_Append_X25519                             = 0.71
+	Weight_CurveIDs_Append_CurveP521                          = 0.46
+	Weight_Extensions_Append_Padding                          = 0.62
+	Weight_Extensions_Append_Status                           = 0.74
+	Weight_Extensions_Append_SCT                              = 0.46
+	Weight_Extensions_Append_Reneg                            = 0.75
+	Weight_Extensions_Append_EMS                              = 0.77
+	Weight_FirstKeyShare_Set_CurveP256                        = 0.25
+	Weight_Extensions_Append_ALPS                             = 0.33
+)
+
 // based on spec's GreaseStyle, GREASE_PLACEHOLDER may be replaced by another GREASE value
 // https://tools.ietf.org/html/draft-ietf-tls-grease-01
 const GREASE_PLACEHOLDER = 0x0a0a
--- a/u_common.go
+++ b/u_common.go
@@ -9,6 +9,7 @@ import (
 	"crypto/sha512"
 	"fmt"
 	"hash"
+	"sync"
 )
 
 // Naming convention:
@@ -104,7 +105,8 @@ type ClientHelloID struct {
 	Client string
 
 	// Version specifies version of a mimicked clients (e.g. browsers).
-	// Not used in randomized, custom handshake, and default Go.
+	// It is used to load weights in func generateRandomizedSpec().
+	// Not used in custom handshake and default Go.
 	Version string
 
 	// Seed is only used for randomized fingerprints to seed PRNG.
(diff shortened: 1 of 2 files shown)


RPRX:
Maybe we should add a field to struct `ClientHelloID` instead, like `Seed`.

gaukas:
Thanks for PR and all the help! 

It looks quite complicated to me and might not be intuitive enough to use. Would you like to add documentation and a few examples?

u_common.go
```diff
@@ -138,6 +140,8 @@ const (
 
 	// versions
 	helloAutoVers = "0"
+
+	DefaultWeights_Version = "0"
```
gaukas, line 144: Is there a specific reason we need it and want it to be exported? 

Also just some styling suggestions, we really don't want a name with underscore(`_`). Possibly name it `DefaultWeightVers` or something.

u_common.go
```diff
@@ -216,6 +220,50 @@ var (
...
+	Weight_Extensions_Append_Reneg                            float64
+	Weight_Extensions_Append_EMS                              float64
+	Weight_OnlyKeyShare_Set_CurveP256                         float64
+	Weight_Extensions_Append_ALPS                             float64
+}
+
+// Please make a copy first instead of modifying them directly.
+var DefaultWeights = Weights{
+	Weight_Extensions_Append_ALPN:                             0.7,
+	Weight_TLSVersMax_Set_VersionTLS13:                        0.4,
```
gaukas, line 250: As discussed if we could update this list of weights it would be great. For example, perhaps it is a good time to have TLS 1.3 to weigh more?

RPRX:
> complicated

Yes, I feels the same. I will extend struct `ClientHelloID` instead.

u_common.go
RPRX: We won't need it anymore.

u_common.go
RPRX: Maybe we should update these preset weights in another PR?

gaukas: Sure. I was about to contact Sergey to learn how did he yield all these parameters.

RPRX:
Upgraded, and fixed a bug introduced by https://github.com/refraction-networking/utls/pull/155.

u_parrots.go
```diff
@@ -2076,11 +2076,11 @@ func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 }
 
 func (uconn *UConn) generateRandomizedSpec() (ClientHelloSpec, error) {
-	return generateRandomizedSpec(uconn.ClientHelloID, uconn.serverName, uconn.HandshakeState.Session, uconn.config.NextProtos)
+	return generateRandomizedSpec(&uconn.ClientHelloID, uconn.serverName, uconn.HandshakeState.Session, uconn.config.NextProtos)
 }
```
RPRX, line 2080: Here.

u_parrots.go
```diff
@@ -2076,11 +2076,11 @@ func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
...
 
 func generateRandomizedSpec(
-	id ClientHelloID,
+	id *ClientHelloID,
```
RPRX, line 2083: Here.

u_parrots.go
```diff
@@ -2076,11 +2076,11 @@ func (uconn *UConn) ApplyPreset(p *ClientHelloSpec) error {
 }
 
 func (uconn *UConn) generateRandomizedSpec() (ClientHelloSpec, error) {
-	return generateRandomizedSpec(uconn.ClientHelloID, uconn.serverName, uconn.HandshakeState.Session, uconn.config.NextProtos)
+	return generateRandomizedSpec(&uconn.ClientHelloID, uconn.serverName, uconn.HandshakeState.Session, uconn.config.NextProtos)
```
RPRX, line 2079: It's expected. https://github.com/refraction-networking/utls#reusing-randomized-fingerprint

u_parrots.go
```diff
@@ -1847,7 +1847,7 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
 	default:
 		if id.Client == helloRandomized || id.Client == helloRandomizedALPN || id.Client == helloRandomizedNoALPN {
 			// Use empty values as they can be filled later by UConn.ApplyPreset or manually.
-			return generateRandomizedSpec(id, "", nil, nil)
+			return generateRandomizedSpec(&id, "", nil, nil)
```
RPRX, line 1850: Actually the original `ClientHelloID` won't be changed, and it's expected too.

u_parrots.go
gaukas: Using pointers here instead may indeed improve some performance, but I wouldn't say it is a bug. 

UTLS is not (yet) optimized for performance anyways... It is better be treated as a prototype instead of product/service. 

And, good work!

gaukas (approved):
This might break someone's code if they are declaring their own ClientHelloIDs using struct literal without fields name (just like us), but it looks good to me otherwise.

gaukas:
If no other things coming to your mind at the moment let's get this merged. We can always make changes later, before the next tagged version (expected in a few weeks).

u_parrots.go
RPRX: It's for https://github.com/refraction-networking/utls#reusing-randomized-fingerprint

RPRX:
It's ready to be merged.

gaukas:
Thanks a ton! Merged.

RPRX:
> This might break someone's code if they are declaring their own ClientHelloIDs using struct literal without fields name

I agree. But declaring variables using an exported struct without fields name is a very bad practice and is expected to be broken.

As v1.2.1 is a little buggy, it's suggested to tag v1.2.2 ASAP once these preset weights are updated (by you, of course).

gaukas:
@RPRX Actually a covert channel between us would be beneficial given that our areas are largely overlapping with each other. 

If you wouldn't mind?

RPRX:
@gaukas I've sent an email to you, please check.

gaukas:
I don't think Sergey is responding to inquiries about uTLS. I will get v1.2.2 released as a hotfix to v1.2.1, then we can discuss a smarter way to assign these default weights.

PR #167, opened 2023-03-03
fix: use badssl.com (TLS 1.2)

### LINKED ISSUE #118: forgeConn() in examples is broken
As mentioned in #117, this issue has been there for awhile. 

## Description
`forgeConn()` in the example in `examples` folder is broken and crashing the example due to null pointer being dereferenced. 

## Cause
The cipher suite checking in `tls.MakeConnWithCompleteHandshake()` doesn't work with **TLS 1.3** yet. 

## Notes
We could find a TLS 1.2 host to demonstrate the example and/or add TLS 1.3 support for `MakeConnWithCompleteHandshake()`.

gaukas:
Partially fixes #118. 
google.com speaks TLS 1.3 which can't be forged in current stage. Use TLS 1.2 instead.

### DIFF
--- a/examples/examples.go
+++ b/examples/examples.go
@@ -285,7 +285,7 @@ func forgeConn() {
 	// then replaces underlying connection of that tls connection with an in-memory pipe
 	// to a forged local in-memory "server-side" connection,
 	// that uses cryptographic parameters passed by a client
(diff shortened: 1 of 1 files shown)


ewust (commented):
It'd be better to fix uTLS to support 1.3

If we go with this, we should leave a comment here detailing the restrictions on what site can be used (e.g. only those that support TLS 1.2)

PR #178, opened 2023-04-01
fix: connection state locked never called

blakebyrnes:
Implication is ALPS peer settings are never copied to the connection state.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1551,6 +1551,9 @@ func (c *Conn) connectionStateLocked() ConnectionState {
 	} else {
 		state.ekm = c.ekm
 	}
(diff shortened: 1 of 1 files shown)


gaukas:
Nice catch @blakebyrnes! I made a mistake when porting the ALPS changes.

blakebyrnes:
Thanks! Do you have a release schedule? Or just a plan for a next one?

gaukas:
> release schedule

We were shooting for monthly release, plus any urgent security fixes. So I would say possibly by the end of this week?

gaukas:
New version v1.3.2 released!

PR #189, opened 2023-05-30
fix build error on loongarch64

yetist:
cpu/cpu_loong64.go:10:7: CacheLineSize redeclared in this block
	cpu/cpu.go:20:5: other declaration of CacheLineSize
cpu/cpu.go:15:30: undefined array length CacheLinePadSize or missing type constraint
cpu/cpu.go:20:29: undefined: CacheLinePadSize
cpu/cpu.go:123:2: undefined: doinit

### DIFF
--- a/cpu/cpu_loong64.go
+++ b/cpu/cpu_loong64.go
@@ -7,7 +7,7 @@
 
 package cpu
 
-const CacheLineSize = 64
+const CacheLinePadSize = 64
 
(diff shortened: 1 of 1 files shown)


gaukas:
We decided to use `golang.org/x/sys/cpu` instead of literally maintaining a fork of it to save everyone's time. A pull request #195 has been opened.

PR #204, opened 2023-07-30
fix: workflow naming convention

gaukas:
Fixes build errors caused by #203.


gaukas:
The error on Go 1.20+ is okay and expected. We are yet to update the test data.

Issue #205, opened 2023-07-31
Zlib decompression fails


felipe-linares:
Seems like a known issue? Tested on facebook.com using the safari TLS.

https://github.com/refraction-networking/utls/pull/95#issuecomment-1190684525

Zlib decompression returns correct `n` but also return `EOF` error.

If you just ignore the error zlib read error it works, but obviously not the ideal scenario.

gaukas:
Hmm. I think currently uTLS supports only `brotli` while the only preset parrot advertising `zlib` is `HelloSafari_16_0`. 

While we would be happy to accept a PR addressing this issue with `zlib` implementation, are there any more recent popular TLS fingerprints advertising `zlib` over `brotli`? (I am guessing Safari?)

gaukas:
Fixed in #206.

PR #206, opened 2023-08-01
Fix zlib cert decompression by ignoring EOF

hwh33:
An io.Reader is allowed (but not required) to return io.EOF upon reading the
last byte of the stream, even if no subsequent bytes were requested. We later
check that we read the expected number of bytes, so we can safely ignore EOF
errors returned by the decompression readers.


hwh33:
Fixes https://github.com/refraction-networking/utls/issues/205

Thanks for calling attention to the issue @yan-foto and @felipe-linares - also for the fix suggestion, which turned out to be correct.

hwh33:
Looks like the same CI failures occur on main right now (and this change should clearly not create any failures as the code path is untested).

go.sum
```diff
@@ -1,26 +1,14 @@
...
-golang.org/x/net v0.7.0 h1:rJrUqqhjsgNp7KqAIc25s9pZnjU7TUcSY7HcVZjdn1g=
-golang.org/x/net v0.7.0/go.mod h1:2Tu9+aMcznHK/AK1HMvgo6xiTLG5rD5rZLDS+rp2Bjs=
 golang.org/x/net v0.11.0 h1:Gi2tvZIJyBtO9SDr1q9h5hEQCp/4L2RQ+ar0qjx2oNU=
 golang.org/x/net v0.11.0/go.mod h1:2L/ixqYpgIVXmeoSA/4Lu7BzTG4KIyPIryS4IsOd1oQ=
-golang.org/x/sys v0.5.0 h1:MUK/U/4lj1t1oPg0HfuXDN/Z1wv31ZJ/YcPiGccS4DU=
-golang.org/x/sys v0.5.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.9.0 h1:KS/R3tvhPqvJvwcKfnBHJwwthS11LRhmM5D59eEXa0s=
 golang.org/x/sys v0.9.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
-golang.org/x/text v0.7.0 h1:4BRB4x83lYWy72KwLD/qYDuTu7q9PjSagHvijDw7cLo=
-golang.org/x/text v0.7.0/go.mod h1:mrYo+phRRbMaCq/xk9113O4dZlRixOauAjOtrjsXDZ8=
```
hwh33, line 24: Drive-by `go mod tidy`.  I'm happy to remove this from the PR though - just let me know.

u_handshake_client.go
```diff
@@ -97,7 +97,7 @@ func (hs *clientHandshakeStateTLS13) decompressCert(m utlsCompressedCertificateM
 	rawMsg[3] = uint8(m.uncompressedLength)
 
 	n, err := decompressed.Read(rawMsg[4:])
-	if err != nil {
+	if err != nil && !errors.Is(err, io.EOF) {
```
hwh33, line 100: As noted in the commit message, we check that `n` equals the expected value below, so we can safely ignore EOF here.

gaukas: Great! Thanks for the clarification!

gaukas:
The CI failure is caused by testdata from Go 1.19, since Go 1.20 updated implementations and essentially broke the old per-byte comparison test. The CI foe Go 1.20 and 1.21 will pass once we bump up version (perhaps when Go 1.21 becomes stable).

gaukas (approved):

Issue #218, opened 2023-08-10
panic on saveSessionTicket


RonaldinhoL:
i just update to latest version and my app crashed

```
func (hs *clientHandshakeState) saveSessionTicket() error {
	if hs.ticket == nil {
		return nil
	}
	c := hs.c

	cacheKey := c.clientSessionCacheKey()
	if cacheKey == "" {
		return nil
	}

	session, err := c.sessionState()
	if err != nil {
		return err
	}
	session.secret = hs.masterSecret

	cs := &ClientSessionState{ticket: hs.ticket, session: session}
	>>>>>    c.config.ClientSessionCache.Put(cacheKey, cs)
	return nil
}
```

line `c.config.ClientSessionCache.Put(cacheKey, cs)`  ClientSessionCache was nil

gaukas:
Hi @RonaldinhoL, 

Will you be able to provide further details on how does your application uses uTLS (especially how you configured your `UClient`/`UConn`). 

The panic was triggered in an unmodified part from our upstream, `crypto/tls`, so will it be possible for you to try connecting to the same server with `crypto/tls` from **Go 1.21.x** and see what happens, while trying your best replicating the configurations. AFAIK, Go 1.21 changed a lot in how session resumptions are handled, so breaking changes are actually expected. 

Please let me know so we could proceed investigating.

gaukas:
Some updates while waiting for your response: 

Seemingly all handshakes leading to a TLS 1.2 connection will fail, and it is not triggered in `utls.Client()` or `crypto/tls.Client()`, thus indeed a bug in our implementation.

gaukas:
So it turns out `crypto/tls` started actively saving session tickets now (seems it wasn't before). If your parrot advertises `session_ticket` extension and the server echoes it back, a session ticket will be received by your client AND it crashes when you don't have a `ClientSessionCache` set. 

A correct use of this new feature is, set a `ClientSessionCache` when you advertises `session_ticket` extension. While I don't see it as a clean fix since it breaks (again) the originality on `crypto/tls` parts, we could definitely push a fix to ignore saving to `ClientSessionCache` when it is nil.

RonaldinhoL:
> So it turns out `crypto/tls` started actively saving session tickets now (seems it wasn't before). If your parrot advertises `session_ticket` extension and the server echoes it back, a session ticket will be received by your client AND it crashes when you don't have a `ClientSessionCache` set.
> 
[...]

PR #220, opened 2023-08-11
retract v1.4.1 panic

trungdlp-wolffun:
Retract v1.4.1 to prevent somebody use this version and get error with `go get github.com/refraction-networking/utls@latest`

Issues: #218 , https://github.com/imroc/req/issues/263


gaukas:
v1.4.1 may not be the only version that needs to be retracted. The actual breaking change has been made in v1.4.0.

go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/refraction-networking/utls
 
 go 1.20
 
+retract (
+	v1.4.1 // #218 panic on saveSessionTicket
```
gaukas, line 6: Perhaps also v1.4.0?

gaukas (approved):
LGTM

trungdlp-wolffun:
@gaukas Sorry, I'm late, but we can add list of versions from to like this: [v1.4.0, v1.4.2], all version between 1.4.0 to 1.4.2 will be retract.

gaukas:
Cool, thanks for the info. I don't think v1.4.2 is problematic so let's keep it for now.

trungdlp-wolffun:
@gaukas Just example haha 🤣

PR #236, opened 2023-08-27
Fix removing session state

own2pwn:
Probably we should check that state is nil rather than the map?


gaukas:
Good catch, I think we missed it there.

PR #244, opened 2023-09-12
fix link to issues

acheong08:
`[issue](issue)` → `https://github.com/refraction-networking/utls/blob/master/issues`
You need to full link


gaukas:
Nice catch... these links are from pre-history era I guess 😄

Issue #246, opened 2023-09-28
[BUG] ? StatusRequestV2Extension


VeNoMouS:
Ran into a situation when using raw bytes, are we sure this was written correctly?

[Shouldn't this](https://github.com/refraction-networking/utls/blob/master/u_tls_extensions.go#L500-L520)

read more like...
```GO
// Write is a no-op for StatusRequestV2Extension. No data for this extension.
func (e *StatusRequestV2Extension) Write(b []byte) (int, error) {
	fullLen := len(b)
	extData := cryptobyte.String(b)
	// RFC 4366, Section 3.6
	var statusType uint8
	var ignored cryptobyte.String
	
	if !extData.ReadUint16LengthPrefixed(&ignored) ||
		!ignored.ReadUint8(&statusType) ||
		!ignored.ReadUint16LengthPrefixed(&ignored) ||
		!ignored.ReadUint16LengthPrefixed(&ignored) {
		//!extData.ReadUint16LengthPrefixed(&ignored) {
		return fullLen, errors.New("unable to read status request v2 extension data")
	}

	if statusType != statusV2TypeOCSP {
		return fullLen, errors.New("status request v2 extension statusType is not statusV2TypeOCSP(2)")
	}

	return fullLen, nil
}
```

`extData` becomes empty after the first `ReadUint16LengthPrefixed`... so you should traverse `ignored` after that...

# Example
```GO
	fmt.Println()
	fmt.Println("Dumping b...")
	fmt.Println(b)
	fmt.Println("Dumping extData")
	fmt.Println(extData)
	fmt.Println()
	
	fmt.Println("Proccessing..")
	fmt.Println(ignored)
	extData.ReadUint16LengthPrefixed(&ignored)
	fmt.Println(ignored)
	ignored.ReadUint8(&statusType)
	fmt.Println(ignored)
	ignored.ReadUint16LengthPrefixed(&ignored)
	fmt.Println(ignored)
	ignored.ReadUint16LengthPrefixed(&ignored) 
	fmt.Println(ignored)
	fmt.Println()
```
```
Dumping b...
[0 7 2 0 4 0 0 0 0]
Dumping extData
[0 7 2 0 4 0 0 0 0]

Proccessing..
[]
[2 0 4 0 0 0 0]
[0 4 0 0 0 0]
[0 0 0 0]
[]
```

gaukas:
Good catch. I agree that this is a bug. Would you love to submit a pull request addressing this issue? 

[...]

PR #247, opened 2023-09-28
fix: sanity check status request v2 extension data (#246)

VeNoMouS:


u_tls_extensions.go
```diff
@@ -503,12 +503,10 @@ func (e *StatusRequestV2Extension) Write(b []byte) (int, error) {
...
 	var statusType uint8
-	var ignored cryptobyte.String
-	if !extData.ReadUint16LengthPrefixed(&ignored) ||
+	if !extData.ReadUint16LengthPrefixed(&extData) ||
 		!extData.ReadUint8(&statusType) ||
-		!extData.ReadUint16LengthPrefixed(&ignored) ||
-		!extData.ReadUint16LengthPrefixed(&ignored) ||
-		!extData.ReadUint16LengthPrefixed(&ignored) {
+		!extData.ReadUint16LengthPrefixed(&extData) ||
+		!extData.ReadUint16LengthPrefixed(&extData) {
```
gaukas, line 509: These two lines are not needed actually since we are not parsing. And the overriding makes it a bit confusing. We should either read them into `&ignored` or just remove these lines.

VeNoMouS: I got rid of `ignore` fwiw since it wasnt used anywhere and just let cryptobyte overwrite `extData` on itself

VeNoMouS: That said, what would you prefer? want me to just remove those 2 checks?

gaukas: It is just a notation that we are intentionally discarding those bytes. uTLS is not optimized for resource usage nor intended for performance-max anyways... 

But if you do care, I would suggest adding a comment that we are not using any of the other bytes, so that other contributors could easily understand what we are doing rather than seeing this as a potential typo/bug. (Especially the last `ReadUint16LengthPrefixed` which caused data loss)

gaukas (approved):

gaukas:
Thanks!

Issue #248, opened 2023-09-30
[BUG] (Fake|Utls)PreSharedKeyExtension


VeNoMouS:
I noticed while trying to implement / understand implementing a PSK for `UtlsPreSharedKeyExtension`  on `ClientHelloSpec`  i ran into a weird problem... so i  started looking at full raw bytes from wireshark for a full client hello replication , the PSK is not actually populated and errors out , im not sure if this due to the recent change with fakePSK to `UtlsPreSharedKeyExtension`

`empty psk detected; remove the psk extension for this connection or set OmitEmptyPsk to true to conceal it in utls`

`
1603010200010001fc0303cee3cca75d5c385f42d7389f48cf75535b403d58159fecdfdf8ef276bf97deba20a581ab3a59648aead9f3a8715e523c82bee686636cf76c50226295badba9e2ee0020aaaa130113021303c02bc02fc02cc030cca9cca8c013c014009c009d002f003501000193aaaa0000000500050100000000ff0100010000230000000b00020100000d0012001004030804040105030805050108060601001200000010000e000c02683208687474702f312e310033002b00299a9a000100001d00203fcef7f05620d78f9aee265cee94d71e0ced2e692532c18f7cdb6add0055a22d446900050003026832001b000302000200170000000a000a00089a9a001d00170018002d0002010100000010000e00000b746c732e706565742e7773002b000706fafa03040303caca0001000015002c00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000029009c00770071854aabbb2c615e18010431cde6cd36b8e5184aab74d3f8e663befea450d68633ad4fbdf8d14a1e2790e04a3434806eea98e7a33fe5c2c5e74c44dba77c7b1da3e475b9fd6345eb5ad557002f95f36f771a40736de379b4cfe741d351d4121963f35abbd4b4c1a2c1c25e234ab30cb2b786857cf5e6002120ff98ad8e80d2b6d6d2f3e3ab73a5447afe9b79cb9e143aaac613b671b1910d58
`

This occurs if `RealPSKResumption` is true or false

VeNoMouS:
ah....

```GO
func (e *UtlsPreSharedKeyExtension) Write(b []byte) (int, error) {
	return len(b), nil // ignore the data
}
```

VeNoMouS:
[...]

PR #251, opened 2023-10-09
fix: Reuse existing PreSharedKeyExtension bug (#248)

VeNoMouS:


gaukas (approved):
Looks good to me! 

Have you tested and are these everything needed to get the bug fixed? Please double check to make sure the [examples](https://github.com/refraction-networking/utls/tree/master/examples/tls-resumption) for session resumption works.

VeNoMouS:
> Looks good to me!
> 
> Have you tested and are these everything needed to get the bug fixed? Please double check to make sure the [examples](https://github.com/refraction-networking/utls/tree/master/examples/tls-resumption) for session resumption works.

Was working in my dev, but let me double check from a clean pull, so all my dev debugging etc from the other week is gone

VeNoMouS:
All appears to be working as should, raw bytes && tls-resumption example

gaukas:
Thanks!

PR #253, opened 2023-10-10
improvement: cleanup

VeNoMouS:
u_common.go:
* Move if statement into existing switch case
* Remove comment about keepPSK

u_fingerprinter.go:
* Remove comment about KeepPSK


gaukas (approved):
Looks good. Thanks!

PR #256, opened 2023-10-13
Deprecate usage of `OmitEmptyPsk` field in `PreSharedKeyExtension` (closes #255)

### LINKED ISSUE #255: Unable to set `OmitEmptyPsk` in `PreSharedKeyExtension`
Not sure if this is considered a bug, but setting `OmitEmptyPsk` to `true` in the `PreSharedKeyExtension` as part of a custom `ClientHelloSpec` has no effect:

```go
spec := &utls.ClientHelloSpec{
    Extensions: []utls.TLSExtension{
        // ...
        &utls.UtlsPreSharedKeyExtension{
            OmitEmptyPsk: true,
        },
    },
}
```

The only way to actually enable `OmitEmptyPsk` is by specifying it in the `UConn` config:

```go
tlsConfig := &utls.Config{OmitEmptyPsk: true}
conn := utls.UClient(rawConn, tlsConfig, tls.HelloCustom)
// ...
```

Because it's overwritten in: 

https://github.com/refraction-networking/utls/blob/1880d7866972add58bdaf1023135200dcbed4f8f/u_session_controller.go#L291

I think this can be fixed/clarified by either making this field private in the extension or by following the configuration in both parts of the code.

Edit: At the time of writing, I'm using the `master` branch with commit 3546fd9286203c4188b0e3f2f01ae7b047552fb5 as the latest change.

sleeyax:
As discussed in the issue.

### DIFF
--- a/u_pre_shared_key.go
+++ b/u_pre_shared_key.go
@@ -134,6 +134,7 @@ type UtlsPreSharedKeyExtension struct {
 	PreSharedKeyCommon
 	cipherSuite  *cipherSuiteTLS13
 	cachedLength *int
+	// Deprecated: Set OmitEmptyPsk in Config instead.
 	OmitEmptyPsk bool
 }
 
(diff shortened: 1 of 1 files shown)


gaukas (approved):
Looks good, thanks!

PR #296, opened 2024-04-29
Update Firefox 120 parrot to a more popular version

adotkhan:
Fixes two missing extension from the HelloFirefox_120 parrot:
* An empty `session_ticket (35)`
* `psk_key_exchange_modes (45)` with `psk_dhe_ke` mode.

This is tested with Firefox 120 and Firefox 123.


adotkhan:
Great, thank you!

PR #297, opened 2024-05-03
quic: always use empty session ID

gaukas:
by RFC 9000 Section 8.4, QUIC must not send non-empty (legacy) session ID, otherwise server will return error.

* It is also weird that so far we see no error from server due to uquic sending 32-byte session ID.


sippejw (approved):
This looks good to me, thanks for fixing this @gaukas!

PR #303, opened 2024-07-16
Fix Config.InsecureSkipTimeVerify not being respected

adotkhan:
This PR fixes an issue where server cert expiry is verified even though `Config.InsecureSkipTimeVerify` is set to true.

It seems that this issue has come about after the uTLS sync with Go 1.21, where the upstream crypto/tls has moved the code to verify server cert expiry above the InsecureSkipVerify check.


gaukas (approved):
Thanks for the nice catch! Currently uTLS has inadequate test coverage which leads to bug risks as such.

gaukas:
If no more edits are going to be made on this pull request, I will be merging it by the end of day. Let me know before that if you want me to hold @adotkhan.

adotkhan:
Thanks for the quick review.

PR #306, opened 2024-07-18
fix: generate ClientHelloSpec only once

adotkhan:
This fixes an issue introduced in https://github.com/refraction-networking/utls/pull/301 where calling `BuildHanshakeStateWithoutSession` with a random profile, generates a new `CilentHelloSpec` on each call.

This PR assume that `UConn.ClientHelloID` is read-only and that it's value should not be modified, which is my understanding of uTLS currently operates. If this assumption is false, another workaround needs to be found.

A review of this is very much appreciated.

### DIFF
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -2588,25 +2588,32 @@ func ShuffleChromeTLSExtensions(exts []TLSExtension) []TLSExtension {
 }
 
 func (uconn *UConn) applyPresetByID(id ClientHelloID) (err error) {
-	var spec ClientHelloSpec
-	uconn.ClientHelloID = id
-	// choose/generate the spec
-	switch id.Client {
-	case helloRandomized, helloRandomizedNoALPN, helloRandomizedALPN:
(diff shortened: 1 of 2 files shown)


u_parrots.go
```diff
@@ -2588,25 +2588,32 @@ func ShuffleChromeTLSExtensions(exts []TLSExtension) []TLSExtension {
...
-		}
-	case helloCustom:
-		return nil
-	default:
-		spec, err = UTLSIdToSpec(id)
-		if err != nil {
-			return err
+
+	if uconn.clientHelloSpec == nil {
+
```
gaukas, line 2593: Just a trivial nit. Could be personal coding style related so feel free to ignore it. 

```suggestion

```

gaukas (approved):
Good catch! Thanks for doing this. I am approving this PR for now, and if no further major change is made, you can expect it get merged by the end of this week.

adotkhan:
Thank you for the review, change applied.

PR #307, opened 2024-07-18
fix: extMasterSecret mismatch with extended_master_secret extension

adotkhan:
Fixes an issue where `clientHelloMsg.extendedMasterSecret`/`PubClientHelloMsg.Ems` is always true, even if the extended_master_secret extension is not set in the Client Hello (e.g. from a random TLS profile).


gaukas (approved):
Thanks for the fix! This change looks good as it should be the responsibility of `(*ExtendedMasterSecretExtension).writeToUConn(uc *UConn)` to set this field.

PR #323, opened 2025-01-13
Merge changes from go 1.23.4

mingyech:
Go 1.23 added native support for X25519Kyber768Draft00. This PR replaces kyber implemented by cloudflare/go to standard library.


noorjs:
They also added ECH for client (I think server implementation coming in 1.24) 😃

mingyech:
@noorjs Thank you for pointing this out. I think utls should support ECH due to its relevance in censorship circumvention. I took a look and I think I'll break this up to a separate PR for ECH because it requires more substantial changes in our client hello marshaller. For now I have made sure it at least works using the `HelloGolang` fingerprint, which uses the go default marshaller.

u_public.go
```diff
@@ -38,17 +41,20 @@ type PubClientHandshakeState struct {
 
 // TLS 1.3 only
 type TLS13OnlyState struct {
-	Suite           *PubCipherSuiteTLS13
+	// Deprecated: Use KeyShareKeys instead.
 	EcdheKey        *ecdh.PrivateKey
 	KeySharesParams *KeySharesParameters
```
ewust, line 46: Should we be removing these old versions? If not (for backward compatibility), we should document here what happens when both KeySharePrivateKeys and these deprecated ones are provided.

u_public.go
```diff
@@ -92,8 +163,9 @@ func (chs13 *clientHandshakeStateTLS13) toPublic13() *PubClientHandshakeState {
 	} else {
 		tls13State := TLS13OnlyState{
 			KeySharesParams: chs13.keySharesParams,
-			EcdheKey:        chs13.ecdheKey,
-			KEMKey:          chs13.kemKey.ToPublic(),
+			EcdheKey:        chs13.keyShareKeys.ecdhe,
+			KEMKey:          chs13.keyShareKeys.publicKEMKey(),
+			KeyShareKeys:    chs13.keyShareKeys.ToPublic(),
```
ewust, line 168: Do we need to provide EchdeKey and KEMKey if KeyShareKeys is provided? If not we should only provide one

u_public.go
```diff
@@ -168,7 +240,9 @@ func (chs12 *clientHandshakeState) toPublic12() *PubClientHandshakeState {
 // }
 
 type CertificateRequestMsgTLS13 struct {
-	Raw                              []byte
+	// Deprecated: crypto/tls no longer use this variable
```
ewust, line 243: Should we delete it? It looks like we still use it, does it have no effect?

u_public.go
```diff
@@ -180,8 +254,13 @@ func (crm *certificateRequestMsgTLS13) toPublic() *CertificateRequestMsgTLS13 {
 	if crm == nil {
 		return nil
 	} else {
+		rawBytes := []byte{}
+		if raw, err := crm.marshal(); err != nil {
```
ewust, line 258: Should we do something with this err? It looks like here we're just silently failing to include the rawBytes if the crm.marshal() fails. Are there errors we might have to catch here?

u_public.go
```diff
@@ -327,7 +405,7 @@ func (shm *serverHelloMsg) getPublicPtr() *PubServerHelloMsg {
 }
 
 type PubClientHelloMsg struct {
-	Raw                          []byte
+	Raw                          []byte // renamed to serverHelloMsg.original in crypto/tls
```
ewust, line 408: renamed to clientHelloMsg.original?

handshake_messages.go
```diff
@@ -148,36 +126,14 @@ func (m *clientHelloMsg) marshal() ([]byte, error) {
 			})
 		})
 	}
-	if m.ticketSupported {
+	if m.ticketSupported && !echInner {
```
ewust, line 129: We may want to provide full control for the user over the outer client hello, even if they're providing an ECH inner. E.g. maybe they want a ticketSupported extension in the outer client hello, and in the inner?

ewust: Maybe ignore this comment :)

ewust (commented):
Just a few comments

u_public.go
mingyech: Good point! I added some docs about when both are set.

u_public.go
mingyech: EchdeKey and KEMKey removed.

u_public.go
mingyech: Added some more docs on how it is not used. It's only not deleted for backwards compatibility, but I doubt anyone is actually using it so it could go either way.

u_public.go
mingyech: The point of generating this rawBytes is to add it to the now deprecated `Raw` field for backwards compatibility (see prev comment). This raw field is not used in utls, so I think we might be able to get away with ignoring this error (and only providing the raw field when there is no error).

u_public.go
mingyech: Good catch!

handshake_messages.go
mingyech: Good point for the ech PR, we will want to provide some way for the user to control wether an extension should be included in the ECH inner or not.

Issue #330, opened 2025-03-04
How to implement the "application_settings (17613)," extended


wangluozhe:
How to implement the "application_settings (17613)," extended.
The previous 17513 has been changed to "application_dettings_old (17513)".

xujinzheng:
just extend the orignal utls.ApplicationSettingsExtension

```go
type ApplicationSettingsExtension struct {
	*utls.ApplicationSettingsExtension
}

func (e *ApplicationSettingsExtension) Len() int {
	bLen := 2 + 2 + 2 // Type + Length + ALPS Extension length
	for _, s := range e.SupportedProtocols {
		bLen += 1 + len(s) // Supported ALPN Length + actual length of protocol
	}
	return bLen
}

func (e *ApplicationSettingsExtension) Read(b []byte) (int, error) {
	if len(b) < e.Len() {
		return 0, io.ErrShortBuffer
	}

	// Read Type.
	b[0] = byte(17613 >> 8)   // hex: 44 dec: 68
	b[1] = byte(17613 & 0xff) // hex: 69 dec: 105

	lengths := b[2:] // get the remaining buffer without Type
	b = b[6:]        // set the buffer to the buffer without Type, Length and ALPS Extension Length (so only the Supported ALPN list remains)

	stringsLength := 0
	for _, s := range e.SupportedProtocols {
		l := len(s)            // Supported ALPN Length
		b[0] = byte(l)         // Supported ALPN Length in bytes hex: 02 dec: 2
		copy(b[1:], s)         // copy the Supported ALPN as bytes to the buffer
		b = b[1+l:]            // set the buffer to the buffer without the Supported ALPN Length and Supported ALPN (so we can continue to the next protocol in this loop)
		stringsLength += 1 + l // Supported ALPN Length (the field itself) + Supported ALPN Length (the value)
	}

	lengths[2] = byte(stringsLength >> 8) // ALPS Extension Length hex: 00 dec: 0
	lengths[3] = byte(stringsLength)      // ALPS Extension Length hex: 03 dec: 3
	stringsLength += 2                    // plus ALPS Extension Length field length
	lengths[0] = byte(stringsLength >> 8) // Length hex:00 dec: 0
	lengths[1] = byte(stringsLength)      // Length hex: 05 dec: 5

	return e.Len(), io.EOF
}

func (e *ApplicationSettingsExtension) UnmarshalJSON(b []byte) error {
	var applicationSettingsSupport struct {
[...]

PR #333, opened 2025-04-01
Implement Chrome 133+ support

BRUHItsABunny:
Since Chrome 133 the TLS fingerprint is changing due to this feature:
https://chromestatus.com/feature/5149147365900288

I found that others are also running into this problem here: https://github.com/refraction-networking/utls/issues/330

It looks like a "full" implementation might be bigger than just extending the current application settings extension, as per another implementation here:
https://github.com/bogdanfinn/utls/commit/fe91add3ecedd324b1b3b395c7b8bc0f72f224da

Opening this PR as a draft to discuss how we should move forward on this, I've also implemented [this](https://github.com/refraction-networking/utls/issues/330#issuecomment-2743313690) for the time being as it might be enough to address this (at least for my use case)

### DIFF
--- a/u_conn.go
+++ b/u_conn.go
@@ -846,9 +846,9 @@ func (c *Conn) utlsConnectionStateLocked(state *ConnectionState) {
 
 type utlsConnExtraFields struct {
 	// Application Settings (ALPS)
-	hasApplicationSettings   bool
-	peerApplicationSettings  []byte
-	localApplicationSettings []byte
+	peerApplicationSettings      []byte
+	localApplicationSettings     []byte
+	applicationSettingsCodepoint uint16
 
 	// Encrypted Client Hello (ECH)
 	echRetryConfigs []ECHConfig
--- a/u_handshake_messages.go
+++ b/u_handshake_messages.go
@@ -54,16 +54,18 @@ func (m *utlsCompressedCertificateMsg) unmarshal(data []byte) bool {
 }
 
 type utlsEncryptedExtensionsMsgExtraFields struct {
-	hasApplicationSettings bool
-	applicationSettings    []byte
-	echRetryConfigs        []ECHConfig
-	customExtension        []byte
+	applicationSettings          []byte
+	applicationSettingsCodepoint uint16
+	echRetryConfigs              []ECHConfig
+	customExtension              []byte
 }
 
 func (m *encryptedExtensionsMsg) utlsUnmarshal(extension uint16, extData cryptobyte.String) bool {
 	switch extension {
 	case utlsExtensionApplicationSettings:
-		m.utls.hasApplicationSettings = true
+		fallthrough
+	case utlsExtensionApplicationSettingsNew:
+		m.utls.applicationSettingsCodepoint = extension
 		m.utls.applicationSettings = []byte(extData)
 	case utlsExtensionECH:
 		var err error
@@ -76,10 +78,10 @@ func (m *encryptedExtensionsMsg) utlsUnmarshal(extension uint16, extData cryptob
 }
 
 type utlsClientEncryptedExtensionsMsg struct {
-	raw                    []byte
-	applicationSettings    []byte
-	hasApplicationSettings bool
-	customExtension        []byte
+	raw                          []byte
+	applicationSettings          []byte
+	applicationSettingsCodepoint uint16
+	customExtension              []byte
 }
 
 func (m *utlsClientEncryptedExtensionsMsg) marshal() (x []byte, err error) {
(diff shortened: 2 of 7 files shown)


Juktong:
Since the previous codepoints are coded in file [exttype_values.go](https://github.com/refraction-networking/utls/blob/master/dicttls/exttype_values.go), it would be a good idea to also update it as well.

BRUHItsABunny:
I noticed some TLS errors especially on Google domains while using my branch. I have updated it to support the new application settings code point in the ECH as well to fix those errors.

u_tls_extensions.go
```diff
@@ -766,6 +768,86 @@ func (e *ApplicationSettingsExtension) Write(b []byte) (int, error) {
 	return fullLen, nil
 }
 
+// ApplicationSettingsExtensionNew represents the TLS ALPS codepoint extension introduced by Chrome 133.
+// More information can be found here: https://chromestatus.com/feature/5149147365900288
+// TODO: This probably should be implemented differently
+type ApplicationSettingsExtensionNew struct {
```
mingyech, line 774: Assuming only the codepoint changed, maybe create an unexported version of `ApplicationSettingsExtension` with a codepoint variable and use it in `Read`. Then the two exported extensions can be embedded it, only changing the codepoint variable before calling `Read`. This way we don't have to keep two versions of the same extension.

u_conn.go
```diff
@@ -749,9 +749,10 @@ func (c *Conn) utlsConnectionStateLocked(state *ConnectionState) {
 
 type utlsConnExtraFields struct {
 	// Application Settings (ALPS)
-	hasApplicationSettings   bool
-	peerApplicationSettings  []byte
-	localApplicationSettings []byte
+	hasApplicationSettings       bool
+	peerApplicationSettings      []byte
+	localApplicationSettings     []byte
+	applicationSettingsCodepoint uint16
```
mingyech, line 755: Any chance we can use `applicationSettingsCodepoint != 0` as `hasApplicationSettings` and remove it?

u_handshake_messages.go
```diff
@@ -54,15 +54,19 @@ func (m *utlsCompressedCertificateMsg) unmarshal(data []byte) bool {
 }
 
 type utlsEncryptedExtensionsMsgExtraFields struct {
-	hasApplicationSettings bool
-	applicationSettings    []byte
-	echRetryConfigs        []ECHConfig
-	customExtension        []byte
+	hasApplicationSettings       bool
```
mingyech, line 57: Same here

u_handshake_messages.go
```diff
@@ -76,10 +80,11 @@ func (m *encryptedExtensionsMsg) utlsUnmarshal(extension uint16, extData cryptob
...
 
 type utlsClientEncryptedExtensionsMsg struct {
-	raw                    []byte
-	applicationSettings    []byte
-	hasApplicationSettings bool
-	customExtension        []byte
+	raw                          []byte
+	applicationSettings          []byte
+	applicationSettingsCodepoint uint16
+	hasApplicationSettings       bool
```
mingyech, line 86: Here as well

mingyech (commented):
Looks good, added some suggestions

RonaldinhoL:
rename ApplicationSettingsExtension to applicationSettingsExtension break my exist custom spec

RonaldinhoL:
Just duplicating the ApplicationSettingsExtension would be cleaner. These changes force unnecessary updates to too many settings.

mingyech:
Ooo yeah we should be careful with compatibility, thanks for pointing out @RonaldinhoL 

@BRUHItsABunny This new change looks good though, looks like nothing will be broken

mingyech:
I don't think it will require outside changes though? applicationSettingsExtension will be auto initialized with the exported versions

BRUHItsABunny:
Could you provide an example of what you are trying to do?
Here is an example of what broke for me (which was copied from somewhere else so I expect it to be more prevalent):
https://github.com/BRUHItsABunny/gOkHttp-ja3spoof/blob/82382ec2ed1f3d51e50cb4479b0142e92d8ee911/transportv2.go#L123

With the new changes this doesn't panic anymore.
Is there any reason you are trying to instantiate the unexported field? It's not a pointer so when you instantiate the outer struct `utls.ApplicationSettingsExtension` it should automatically make it.

mingyech:
Added 4241c5b6556cb44f7694d2289d1e2071c1bbae2f to make clear that this will not break backward compatibility, but let me know if there's something else broken by this

> btw if u want to merge to master, you'd better make a squash commit.

Good idea yeah I'll do some squashing when merging

BRUHItsABunny:
Thanks for carrying it over the finish line @mingyech

Issue #345, opened 2025-05-15
Signature verification error visiting URL


BRUHItsABunny:
When I visit this URL [https://cards-na.klarna.com](https://cards-na.klarna.com) I get the following error:
`tls: invalid signature by the server certificate: crypto/rsa: verification error`

I tracked it to [this](https://github.com/refraction-networking/utls/blob/dbca939a468f603cf8b96c50e7735be153b77233/handshake_client_tls13.go#L869) and [this](https://github.com/refraction-networking/utls/blob/dbca939a468f603cf8b96c50e7735be153b77233/auth.go#L57) but that leaves me pretty unsure on why this occurs given the TLS connection is fine in my browser (Brave) as well as std lib Go.

Replicate (modified from the [old example](https://github.com/refraction-networking/utls/blob/master/examples/old/examples.go) in the repo):
```go
package main

import (
	"bufio"
	"fmt"
	"net"
	"net/http"
	"net/http/httputil"
	"net/url"
	"time"

	tls "github.com/refraction-networking/utls"
	"golang.org/x/net/http2"
)

var (
	dialTimeout = time.Duration(15) * time.Second
)

// var requestHostname = "facebook.com" // speaks http2 and TLS 1.3
// var requestAddr = "31.13.72.36:443"
var requestHostname = "cards-na.klarna.com"
var requestAddr = "cards-na.klarna.com:443"

func HttpGetDefault(hostname string, addr string) (*http.Response, error) {
	config := tls.Config{ServerName: hostname}
	dialConn, err := net.DialTimeout("tcp", addr, dialTimeout)
	if err != nil {
		return nil, fmt.Errorf("net.DialTimeout error: %+v", err)
	}
	tlsConn := tls.Client(dialConn, &config)
	defer tlsConn.Close()
	return httpGetOverConn(tlsConn, tlsConn.ConnectionState().NegotiatedProtocol)
}

func HttpGetStdLib(hostname string, addr string) (*http.Response, error) {
	return http.Get("https://" + hostname + "/")
}

func HttpGetByHelloID(hostname string, addr string, helloID tls.ClientHelloID) (*http.Response, error) {
	config := tls.Config{ServerName: hostname}
	dialConn, err := net.DialTimeout("tcp", addr, dialTimeout)
	if err != nil {
		return nil, fmt.Errorf("net.DialTimeout error: %+v", err)
	}
	uTlsConn := tls.UClient(dialConn, &config, helloID)
	defer uTlsConn.Close()

[...]

PR #348, opened 2025-06-01
[Cleaner Version] Support for Certificate Compression Algorithm extension in Certificate Request Message

Juktong:
Intended to do the same functionality as my previous pr #347  but for a cleaner commit history. Hence the following description is copied from previous pr.

---

This version introduces a new CertCompressionAlgo implementation for certificateRequestMsgTLS13 in handshake_messages.go, addressing issue https://github.com/refraction-networking/utls/issues/345.

The change resolves the error: 'tls: invalid signature by the server certificate: crypto/rsa: verification error'
which occurs when connecting to servers that include a certificate compression algorithm extension in the CertificateRequest message during the TLS 1.3 handshake.

@BRUHItsABunny — would you mind testing this on your end to confirm the fix works as expected?

### DIFF
--- a/handshake_messages.go
+++ b/handshake_messages.go
@@ -1280,6 +1280,7 @@ type certificateRequestMsgTLS13 struct {
 	scts                             bool
 	supportedSignatureAlgorithms     []SignatureScheme
 	supportedSignatureAlgorithmsCert []SignatureScheme
+	certRequestCompressionAlgs       []CertCompressionAlgo
 	certificateAuthorities           [][]byte
 }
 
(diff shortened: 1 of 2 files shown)


handshake_messages.go
```diff
@@ -1276,10 +1276,12 @@ func (m *newSessionTicketMsgTLS13) unmarshal(data []byte) bool {
 }
 
 type certificateRequestMsgTLS13 struct {
+	original                         []byte
```
mingyech, line 1279: In non u_\* files let's comment changes to distinguish between utls and stdlib code, it makes it easier when merging with upstream. See [here](https://github.com/refraction-networking/utls/blob/c567269913b5e46f670ec8eeebc16a6ab9fc41d4/handshake_client_tls13.go#L754-L757) and [here](https://github.com/refraction-networking/utls/blob/c567269913b5e46f670ec8eeebc16a6ab9fc41d4/handshake_client_tls13.go#L46)

examples/ech/sslkeylogging.log
mingyech, line 1: untrack this?

handshake_messages.go
```diff
@@ -1276,10 +1276,12 @@ func (m *newSessionTicketMsgTLS13) unmarshal(data []byte) bool {
...
 	ocspStapling                     bool
 	scts                             bool
 	supportedSignatureAlgorithms     []SignatureScheme
 	supportedSignatureAlgorithmsCert []SignatureScheme
+	certRequestCompressionAlgs       []CertCompressionAlgo
```
mingyech, line 1284: It looks like you also have your second approach of fixing this from https://github.com/refraction-networking/utls/issues/345#issuecomment-2927991151 in here as well with the `original` and `originalBytes()`, could we remove the first approach if it works without it? It's a good reference in case we need to actually implement this in the future but I think your new approach is better.

mingyech (commented):
Thank you for contributing! It looks good to me, just have a few nit picks.

PR #353, opened 2025-07-02
Update u_parrots.go

yuhan6665:
Looks like a typo that "TLS_AES_256_GCM_SHA384" is defined twice for Edge_106


mingyech (approved):
Good catch, thanks!

Issue #359, opened 2025-07-24
`PubServerHelloMsg.ServerShare` is not exported correctly


wwqgtxx:
As the title says, `PubServerHelloMsg.ServerShare` uses the unexported `keyShare` type instead of `KeyShare`, which makes it impossible for external programs to access internal fields in a safe way.

https://github.com/refraction-networking/utls/blob/ab3777592f7547da3ec36ca5983223977e51b498/u_public.go#L287

RPRX:
~~昨天看到这个 issue 的时候还没想到是和 https://github.com/XTLS/Xray-core/pull/4929 相关的~~

RPRX:
@wwqgtxx ~~我感觉这可能需要你开个 PR 了~~

PR #361, opened 2025-07-26
fix: `PubServerHelloMsg.ServerShare` is not exported correctly

wwqgtxx:
ref: https://github.com/refraction-networking/utls/issues/359


mingyech (approved):
Thanks for the fix!

PR #365, opened 2025-08-05
feat: Config.NextProtos for inner CH when ECH is enabled

### LINKED ISSUE #355: [feature request] Hide alpn in ECH
Like this: Outer(http/1.1+h2) innter(http/1.1) to hide we are tring to negotiate http/1.1

For a long time, I was thinking about how to design usage
One of my thought is:
For outer: Always use http/1.1+h2 when using browser fingerprint
For inner: Use Config.NextProtos
Current behavior of utls(outer and inner) is like "for outer" above(and completely ignore Config.NextProtos)
so there is only one thing to change:when building inner hello, use Config.NextProtos to override copied alpn from outer

Juktong:
Fixes #355 by letting `Config.NextProtos` determine the ALPN list in the inner ClientHello when ECH is enabled, while the outer ClientHello continues to use the preset ALPN from `ClientHelloSpec.Extensions`.

A unit test for this ALPN-ECH will be submitted in another commit.

### DIFF
--- a/u_tls_extensions.go
+++ b/u_tls_extensions.go
@@ -611,8 +611,11 @@ type ALPNExtension struct {
 }
 
 func (e *ALPNExtension) writeToUConn(uc *UConn) error {
-	uc.config.NextProtos = e.AlpnProtocols
-	uc.HandshakeState.Hello.AlpnProtocols = e.AlpnProtocols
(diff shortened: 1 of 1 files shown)


mingyech:
Closing in favor of #383. This was a good attempt

PR #371, opened 2025-09-23
fix: darwin arm64 always has AESGCMHardwareSupport

wwqgtxx:
While merging the relevant commits for go1.25, we discovered a minor issue: the value of `hasAESGCMHardwareSupport` was always false on ARM64 Macs.

Comparing the values ​​of `cpu.ARM64.HasAES` and `cpu.ARM64.HasPMULL` in `internal/cpu` and `x/sys/cpu` confirmed inconsistencies.

Given that `hasAESGCMHardwareSupport` is heavily used in TLS, maintaining a consistent value with the standard library ensures synchronization with the standard library's behavior.

From the comments in the standard library: https://github.com/golang/go/blob/go1.25.0/src/internal/cpu/cpu_arm64_darwin.go#L26-L30

For ARM64 Darwin platforms (excluding iOS), AES GCM hardware acceleration is always available.


wwqgtxx:
This PR has been superseded by #372, which has a better implementation.

PR #372, opened 2025-09-23
replace `x/sys/cpu` to embed `internal/cpu`

wwqgtxx:
While merging the relevant commits from go1.25, we discovered a minor issue: the value of `hasAESGCMHardwareSupport` was consistently false on arm64 Macs.

Comparing the values ​​of `cpu.ARM64.HasAES` and `cpu.ARM64.HasPMULL` in `internal/cpu` and `x/sys/cpu` confirmed inconsistencies.

Extensive testing revealed that the values ​​of `cpu.ARM64.HasAES` and `cpu.ARM64.HasPMULL` were consistently false on arm64 Darwin and Android platforms.

Given the extensive use of `hasAESGCMHardwareSupport` in tls, we embedded `internal/cpu` in go1.25.0 to ensure consistent behavior with the standard library.

### DIFF
--- a/cipher_suites.go
+++ b/cipher_suites.go
@@ -19,7 +19,7 @@ import (
 	_ "unsafe" // for linkname
 
 	"github.com/refraction-networking/utls/internal/boring"
-	"golang.org/x/sys/cpu"
+	"github.com/refraction-networking/utls/internal/cpu"
 
 	"golang.org/x/crypto/chacha20poly1305"
 )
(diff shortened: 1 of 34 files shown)


mingyech:
Thanks for the PR, and sorry for the late reply. It looks like that issue has been fixed at https://github.com/golang/sys/commit/3687fbd71652878ab091f7272b84537b63fe0b55 so I'm going to wait for that to tag a new version and updating it instead of adding the code here to avoid introducing another thing to keep up to date. This was a helpful insight.

PR #380, opened 2026-01-13
feat: add Yandex Browser fingerprint (HelloYandex_Auto)

hxehex:
added HelloYandex_Auto fingerprint based on Yandex Browser for Windows 10 x86_64

updated based on deep traffic analysis:
uses ShuffleChromeTLSExtensions to match real randomization behavior
added SessionTicketExtension and UtlsPreSharedKeyExtension for warm handshakes (resumption)
pq key share X25519MLKEM768 (0x11ec)
applicationsettings (ALPS) extension with correct ID (0x44cd) via ApplicationSettingsExtensionNew
brotli certificate compression

### DIFF
--- a/u_parrots.go
+++ b/u_parrots.go
@@ -976,7 +976,7 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
 			CompressionMethods: []byte{
 				0x00, // compressionNone
 			},
-			Extensions: []TLSExtension{
+			Extensions: ShuffleChromeTLSExtensions([]TLSExtension{
 				&UtlsGREASEExtension{},
 				&StatusRequestExtension{},
 				&PSKKeyExchangeModesExtension{[]uint8{
@@ -993,6 +993,7 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
 					PSSWithSHA512,
 					PKCS1WithSHA512,
 				}},
+				&SessionTicketExtension{},
 				&ALPNExtension{AlpnProtocols: []string{"h2", "http/1.1"}},
 				&ExtendedMasterSecretExtension{},
 				&SupportedVersionsExtension{[]uint16{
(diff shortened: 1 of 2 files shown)


hxehex:
BUMP

hxehex:
...?

hxehex:
hi

u_parrots.go
```diff
@@ -953,6 +953,80 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
 				&UtlsGREASEExtension{},
 			}),
 		}, nil
+	case HelloYandex_Auto:
```
mingyech: Should have the browser version here instead of auto.

u_parrots.go
```diff
@@ -953,6 +953,80 @@ func utlsIdToSpec(id ClientHelloID) (ClientHelloSpec, error) {
...
+				&SCTExtension{},
+				&SupportedPointsExtension{SupportedPoints: []byte{
+					0x00, // pointFormatUncompressed
+				}},
+				BoringGREASEECH(),
+				&UtlsCompressCertExtension{[]CertCompressionAlgo{
+					CertCompressionBrotli,
+				}},
+				&UtlsGREASEExtension{},
+				&UtlsPreSharedKeyExtension{},
```
mingyech: The PSK extension should be removed as it is not sent in new connections. In most cases the PSK extension is used to resume previous TLS sessions, so you might see it when you visit a previously opened website. utls should be mimicking new connections similar to other fingerprints in this file.

mingyech (commented):
Thanks for the PR, and sorry for the delay. It seems like the fingerprint is the same as `HelloChrome_133` after my comment about the PSK extension below (its using chromium so it makes sense). So I think utls should just use that to reduce duplicate code. But I'm not opposed to adding a `HelloYandex_Auto` alias to `HelloChrome_133` to show they are the same and to track future versions. Tested with Yandex 25.12.3.1140.

hxehex:
done
I removed the duplicate code block and set HelloYandex_25_12 as an alias to HelloChrome_133
HelloYandex_Auto is now an alias to it

PR #383, opened 2026-02-23
Hide ALPN in ECH

### LINKED ISSUE #355: [feature request] Hide alpn in ECH

Fangliding:
Close #355

Tested and works perfectly. wss alpn `http/1.1` is hidden.
image

When I finished the work I found #365 but it didn't write ALPN correctly when building the inner hello

### DIFF
--- a/handshake_messages.go
+++ b/handshake_messages.go
@@ -242,10 +242,11 @@ func (m *clientHelloMsg) marshalMsgReorderOuterExts(echInner bool, outerExts []u
 		}
 	}
 	if len(m.alpnProtocols) > 0 {
-		// RFC 7301, Section 3.1
-		if echInner {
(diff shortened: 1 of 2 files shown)


mingyech (commented):
Thanks, merging

PR #386, opened 2026-03-02
Fix race condition in loadSession with shared session cache

thesinakamali:
When multiple goroutines share a ClientSessionCache, loadSession() returns a *SessionState pointer that is shared across connections. Subsequent code (e.g. SessionTicketExtension.InitializeByUtls) mutates this session object, causing a data race. Fix by making a shallow copy of the SessionState immediately after retrieving it from the cache.

### DIFF
--- a/handshake_client.go
+++ b/handshake_client.go
@@ -435,6 +435,11 @@ func (c *Conn) loadSession(hello *clientHelloMsg) (
 	}
 	session = cs.session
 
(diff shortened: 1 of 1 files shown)


maciejkmak:
@thesinakamali man your PR pointed me into good direction when investigating why my server is failing in production. I spent over 60h in total looking for problem.

THANK YOU! Live long and prosper 🖖

to maintainers: please merge it

PR #401, opened 2026-06-24
Add Chrome 133 CNSA ClientHello

Medium1992:
Adds a Chrome 133 CNSA ClientHello variant with AES-256-GCM preferred over AES-128-GCM in the TLS 1.3 cipher suite list.

The rest of the profile matches HelloChrome_133.


Medium1992:
@mingyech @sippejw (cc @ewust)

Is utls still actively maintained? A few things from a downstream user:

- Newer browser fingerprints — Chrome 150+ and the latest Safari. Some Safari
  work already landed on master but hasn't made it into a release yet.

- Post-quantum: ML-KEM key shares are in now — as browsers start advertising
  ML-DSA in signature_algorithms, would those be tracked too, so the presets
  stay accurate?

- Most importantly, a new stable release. The last tag is v1.8.2 (Jan 2026),
  and quite a bit has merged since (ML-KEM, GREASE fixes, newer Safari). Many
  downstream projects pull only tagged stable releases as a
  matter of policy, so everything currently on master is out of reach for them.

Even a short note on the maintenance and release plans would help a lot. Thanks
for the work on this.

PR #409, opened 2026-07-21
Preserve certificateMsgTLS13 original bytes for the TLS 1.3 transcript

sam-at-ava:
## Problem

TLS 1.3 handshakes can fail at `CertificateVerify` with:

```
tls: invalid signature by the server certificate: crypto/rsa: verification error
```

against real production HTTPS endpoints — even though the certificate chain validates and the same server works with `crypto/tls` and with a Chrome fingerprint.

## Root cause

`certificateMsgTLS13` does not implement `handshakeMessageWithOriginalBytes`. When uTLS reads the server Certificate with `readHandshake(nil)` (to support RFC 8879 cert compression) and then hashes it into the transcript via `transcriptMsg`, the fallback path **re-marshals** the parsed message with `marshalCertificate()`:

```go
func transcriptMsg(msg handshakeMessage, h transcriptHash) error {
    if msgWithOrig, ok := msg.(handshakeMessageWithOriginalBytes); ok {
        if orig := msgWithOrig.originalBytes(); orig != nil {
            h.Write(orig)              // faithful
            return nil
        }
    }
    data, err := msg.marshal()         // <- re-marshal, may diverge
    ...
}
```

`marshalCertificate()` is **not guaranteed to reproduce the peer's byte encoding**: it only re-emits OCSP/SCT (in a fixed order) for the leaf, and drops any other per-certificate extension; non-canonical length encodings are also lost. When the server's Certificate message doesn't round-trip exactly, the client's transcript diverges from the server's, so the `CertificateVerify` signature — which is correct — fails to verify against the client's (wrong) transcript hash.

This only affects the **uncompressed**-certificate path; compressed certificates already transcript their raw `CompressedCertificate` bytes. It's deterministic and reproducible against production servers whose leaf Certificate carries an extension beyond OCSP/SCT.

## Fix

`clientHelloMsg`, `serverHelloMsg` and `certificateRequestMsgTLS13` already cache their original wire bytes for exactly this reason (the June-2025 `originalBytes` work). `certificateMsgTLS13` was the one TLS 1.3 transcript message left out. This PR:

- adds `original []byte` to `certificateMsgTLS13`, captures it in `unmarshal`, and returns it from `originalBytes()`;
- nils it in `TestMarshalUnmarshal` alongside the sibling messages;
- adds a regression test that builds a Certificate whose leaf carries an extension `marshalCertificate` drops, and asserts `transcriptMsg` hashes the original wire bytes rather than the divergent re-marshal.

`go test ./...` passes; the new test fails without the `handshake_messages.go` change.

### DIFF
--- a/handshake_messages.go
+++ b/handshake_messages.go
@@ -1499,6 +1499,7 @@ func (m *certificateMsg) unmarshal(data []byte) bool {
 }
 
 type certificateMsgTLS13 struct {
+	original     []byte // [uTLS]
 	certificate  Certificate
 	ocspStapling bool
 	scts         bool
@@ -1561,7 +1562,7 @@ func marshalCertificate(b *cryptobyte.Builder, certificate Certificate) {
 }
 
 func (m *certificateMsgTLS13) unmarshal(data []byte) bool {
-	*m = certificateMsgTLS13{}
+	*m = certificateMsgTLS13{original: data} // [uTLS]
 	s := cryptobyte.String(data)
 
 	var context cryptobyte.String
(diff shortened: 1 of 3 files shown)


hiroTamada (approved):
LGTM

